PR #1, opened 2015-05-26
Init proposal

ahilles107:
add clean symfony 2.7 with docker configuration


takeit:
There should be, in my opinion, the guide on how to run the project using the docker.

Also why are we not using the new Symfony 3 directory structure? were there some issues with that ? from symfony 3 this new directory structure will be available by default and we will need to upgrade the project anyway to this new structure.

ahilles107:
Hmm... from what i know this is Symfony 3 directory structure. 

About docker - there is section ```How to build it with docker:``` it just need to be polished and expanded with next pr's

ahilles107:
ahhh it isn't i forget about that paremeter when i was creating project. Fork my branch and push change as PR please.

composer.json
```diff
@@ -0,0 +1,58 @@
+{
+    "name": "superdesk/web-renderer",
+    "license": "proprietary",
+    "type": "project",
+    "autoload": {
+        "psr-0": {
```
takeit: PSR-0 is depracated. PSR-4 is now recommended as an alternative. (http://www.php-fig.org/psr/psr-0/)

PR #3, opened 2015-05-28
init docs, add lipp/theme remove assetic

ahilles107:


takeit:
shouldnt we keep the documentation in a separate repository for the Sphinx ?

ahilles107:
don't think so. Keeping it in this same don't require additional pull requests, we can sen feature and documentation in one PR. It's easier to keep it in sync. What it can break? Why not keep it here?

takeit:
I just think that the documentaion is not needed here at all because in production those file will just lay in the web renderer directory when installed. People will use the link for the shpinx documentation anyway to read the docs

ahilles107:
We can easy remove it on packaging. There should be also option to exclude it from automatically generated zipball (on github). Keeping doc's close to source makes it easier to maintain. And there no bigger bug that not maintained documentation.

takeit:
:+1:

PR #5, opened 2015-05-29
inatall phpspec, add travis config, add link to online documentation

ahilles107:


.travis.yml
```diff
@@ -0,0 +1,11 @@
+language: php
+php:
+  - 5.6
```
takeit: in composer.json version should be also set to >=5.6 instead >=5.3.3

ahilles107: done

PR #11, opened 2015-06-18
 Updates docker with entrypoint script setup

m038:
- Installs composer dependencies automatically
  - Chowns all command by user on each run
  - Also makes a volume in php-fpm container with web dir

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,18 +36,21 @@ Full documentation can be found here: [http://web-renderer.readthedocs.org/][1]
 
 ## How to build it with docker:
 
-#### Build php and nginx images:
+#### Setup
+
+Add ```127.0.0.1 webrenderer.dev``` to ```/etc/hosts```. **Mac users** use the ip of your virtualbox (e.g. ```boot2docker ip```).
+
+#### Build:
 
 ```bash
-docker build -t webrenderer/php-fpm docker/php-fpm/
-docker build -t webrenderer/nginx docker/nginx/
+docker-compose build
 ```
-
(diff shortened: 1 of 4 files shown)


takeit:
I would hold on with it, I found there some issues. First  of all there is no curl, volumes are not configured properly which actually doesnt mount the host repo to the container repo, it would be better to create a sepearate Dockerfile config for the renderer and separate things from each other

ahilles107:
@takeit - can you explain more?

m038:
Curl is added to the list of packages to install.

takeit:
in the entrypoint file there is installation of the composer using curl, curl is not defined to be installed in the Dockerfile, thus it fails due to missing bin file. Also mapping the volumes is incosistent (but that was introduced in previous PRs). Volumes for the symofny app are defined in the nginx dockerfile - this needs to be done preferably in a separate Dockerfile that will run commands related only to the symfony code and not to the php and nginx

ahilles107:
Yeah, it can be in separate docker - not the nginx one - it was like this on beginning but i wanted to check if performance of docker will be better with just two images instead 3. But looks like on mac boot2docker is an issue.

ahilles107:
@m038 - can you move nginx to seperate container?

m038:
@ahilles107 i think @takeit was already working on this. Let's wait for his PR.

takeit:
please, don't wait for me, had some issues with the docker config for separate container. Seems it needs more time to investigate. As for now, this can be merged.

PR #16, opened 2015-07-28
[WIP] SWP-17 Define and implement first version of Updates Service

takeit:
- [x] - set up API bundles (JMSSerializerBundle, FOSRestBundle, NelmioApiDoc, BazingaHateoasBundle)
- [x] - create guzzle client to connect to external update server
- [x] - create Updates manager to manage updates
- [x] - add service to make use of https://github.com/ahilles107/updater
- [x] - configure logger for separate channel: "updater"
- [x] - rename AppBundle to SWP\WebRendererBundle
- [x] - Move Version class from UpdaterBundle to AppBundle (after rename to WebRendererBundle)
- [x] - write documentation
- [x] - move this bundle to separate repo (atm https://github.com/takeit/SWPUpdaterBundle I will make a PR once I finish tests)

***TODO:***

- [ ] - add tests
- [ ] - rebase


src/SWP/UpdaterBundle/Manager/AbstractManager.php
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    /**
+     * Checks for the latest version from the available ones.
+     *
+     * @param array $versions An array of versions
+     *
+     * @return bool
+     */
+    protected function checkForLatestVersion(array $versions)
```
ahilles107: i would implement special endpoint on server for fetching latest version instead looping here

src/SWP/UpdaterBundle/README.md
```diff
@@ -0,0 +1,2 @@
+# SWPUpdaterBundle
+Provides automatic Symfony2 apps upgrades
```
ahilles107: integrates symfony2 with ahilles107/updater library wutch provide easy way for php applications updates.

src/SWP/UpdaterBundle/Manager/AbstractManager.php
```diff
@@ -0,0 +1,106 @@
...
+    /**
+     * Construct.
+     *
+     * @param ClientInterface $client Update server client
+     */
+    public function __construct(ClientInterface $client)
+    {
+        $this->client = $client;
+        // TODO
+        // inject here Setting service to get the current app version.
```
ahilles107: Maybe we can define some interface for Version class witch need to be passed here (and implemented in checked app) ?

ahilles107: I think that would be better to keep version in class file, not in storage (database).

takeit: yeah I also thought about this, aready created Version model with a proper interface, will need to finish it

src/SWP/UpdaterBundle/Manager/AbstractManager.php
hlmnrmr: :+1:

ahilles107: or at least fetching all available packages from current version (as we will update apps one by one)

app/config/config.yml
```diff
@@ -7,6 +7,7 @@ imports:
 # http://symfony.com/doc/current/best_practices/configuration.html#application-related-configuration
 parameters:
     locale: en
+    swp_updater.version.class: SWP\UpdaterBundle\Version\Version
```
ahilles107: i would move it to services.yml file inside bundle

app/config/config.yml
```diff
@@ -71,3 +73,42 @@ swiftmailer:
...
+            'Symfony\Component\Routing\Exception\ResourceNotFoundException': true
+
+sensio_framework_extra:
+    view:    { annotations: false }
+
+swp_updater:
+    version_class: %swp_updater.version.class%
+    monolog_channel: true
+    client:
+        base_uri: http://rafal.sourcefabric.net
```
ahilles107: this should be:

```
https://updates.sourcefabric.org 
```

you can edit /etc/hosts and add there redirect to your container ip

app/config/config.yml
takeit: this is an example Version class, I just put it here so the UpdaterBundle can work for me.

Do you mean AppBundle or UpdaterBundle ?

app/config/config.yml
takeit: yeap, will do

app/config/config.yml
ahilles107: ahhh... yeah, but then example should not to be committed (in our case we need here our project version class) - in AppBundle (please also rename AppBundle to SWP\WebRendererBundle)

takeit: yes, ok will do

src/SWP/UpdaterBundle/Manager/UpdateManager.php
```diff
@@ -0,0 +1,214 @@
...
+use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
+
+/**
+ * Update manager.
+ */
+class UpdateManager extends AbstractManager
+{
+    const UPDATES_ENDPOINT = '/check.json';
+    const CORE_ENDPOINT = '/core.json';
+    const LATEST_VERSION_ENDPOINT = '/check2.json';
```
ahilles107: why not /latest_version.json ?

takeit: I will change that to a proper naming, these are only test names

src/SWP/UpdaterBundle/Model/UpdatePackage.php
```diff
@@ -0,0 +1,49 @@
...
+ */
+
+namespace SWP\UpdaterBundle\Model;
+
+use Updater\Package\Package;
+
+/**
+ * Update package model class.
+ */
+class UpdatePackage extends Package
```
ahilles107: why not adding this to Updater\Package\Package ?

src/SWP/UpdaterBundle/Resources/config/routing.yml
```diff
@@ -0,0 +1,8 @@
+swp_updater:
+    resource: "@SWPUpdaterBundle/Controller/"
+    type:     annotation
+    prefix:   /
+
+NelmioApiDocBundle:
+    resource: "@NelmioApiDocBundle/Resources/config/routing.yml"
+    prefix:   /api/doc
```
ahilles107: missing empty line at end of file

src/SWP/UpdaterBundle/Resources/translations/messages.fr.xlf
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0"?>
```
ahilles107: not used file - please remove

ahilles107:
:+1:

src/SWP/UpdaterBundle/Model/UpdatePackage.php
takeit: needed to set some extra property (url) which is not used by the ahilles107/updater, but is needed for UpdaterBundle purposes

app/config/config.yml
m038: With Guzzle you can also specify in the options to resolve a url differently. Check the readme of my bundle, it allows to set default options for every request, with the following config you can achieve this.

curl:
    10203: # CURLOPT_RESOLVE
         - "example.com:5050:localhost"  # This will resolve the host example.com to your localhost

src/SWP/UpdaterBundle/Client/RemoteClient.php
```diff
@@ -0,0 +1,63 @@
...
+{
+    private $baseUri;
+
+    /**
+     * {@inheritdoc}
+     */
+    public function __construct(array $config = array())
+    {
+        parent::__construct($config);
+        $this->baseUri = $config['base_uri'];
```
m038: Shouldn't you check if base_uri is set?

m038:
:+1:

app/config/config.yml
ahilles107: i see new class but here is still ld namespace

takeit: yeah I didnt push config.yml yet

app/config/config.yml
```diff
@@ -71,3 +72,28 @@ swiftmailer:
...
+        messages:
+            'Symfony\Component\Routing\Exception\ResourceNotFoundException': true
+
+sensio_framework_extra:
+    view:    { annotations: false }
+
+swp_updater:
+    version_class: %swp_updater.version.class%
+    client:
+        base_uri: http://updates.sourcefabric.org
```
ahilles107: i have problem with naming here: 

is it really client configuration? And base url shouldn't be called updates_server_uri ?

takeit: In updater bundle there is a Guzzle client and you can have your own clients too if you want. This client connects to updates external server, so I called it client, like swp_updater client

ahilles107: ok, but base_uri don't tell me anything - base_uri for what? client? but it's base_uri for updates server.

takeit: I also created it under the `client` so it has some structure and in the future we can add more options if needed, like for example type of client: guzzle or buzz etc. I think `base_uri` name is enough

takeit: yeah its base uri of the updates server, but in the context that we have the structure of clients in the UpdaterBundle.

takeit: @m038 what do you think?

m038: I think it needs to be documented any way so in my opinion base_uri is fine. Any that is more clear is longer, think of update_host_base_uri or update_server_base_uri.

takeit: yeah I also agree

PR #22, opened 2015-08-26
Bridge: Adds configuration and dependency bridge bundle

m038:


app/config/config.yml
```diff
@@ -121,3 +121,6 @@ swp_updater:
     version_class: %swp_updater.version.class%
     client:
         base_uri: http://updates.sourcefabric.org
+
+swp_bridge:
+    base_uri: 'http://example.com:5050'
```
ahilles107: let's change it for ```http://superdesk.dev:5050``` - we will make it configurable later from ui bot for now would be better to just create host like that locally and make it working without modyfing content. Example.com isn't best name in this case.

app/config/config_dev.yml
```diff
@@ -43,3 +43,9 @@ monolog:
 
 #swiftmailer:
 #    delivery_address: me@example.com
+
+swp_bridge:
+    options:
+        curl: # http://guzzle.readthedocs.org/en/latest/faq.html#how-can-i-add-custom-curl-options
+            10203: # integer value of CURLOPT_RESOLVE
+                 - "example.com:5050:localhost"  # This will resolve the host example.com to your localhost
```
ahilles107: let's got wih ```http://superdesk.dev:5050``` also here. and question is if we realy need this in config_dev.yml - as a developer i can easly create host like that. No port forwarding is needed

ahilles107:
:+1:

PR #24, opened 2015-09-14
Storage bundle

m038:
# General responsibility of the bundle
The storage bundle will handle all tasks in regards to storing and retrieving data. It will provide a REST api for interaction with the UI Client for the web publisher. Further more it will also provide an internal api, based on some interfaces, with which it would be possible to change the underlying Doctrine PHPCR ODM library with another storage library.
Finally it will contain the document models for the by default supported document types.

## API Endpoints
There should be endpoints to operate CRUD actions on Documents. The API will follow the following url structure. Next to that is it possible to modify different document types via the API, in the example the document type _Article_ is used.
All output of the API will be in JSON.

Method | URL
---------- | ------
GET |  /api/articles/{id} 
POST |  /api/articles/
PUT |  /api/articles/{id} 
DELETE | /api/articles/{id}

The API bundle (yet to be defined) should handle all responsibility for matching routes to controllers and should provide us with the needed tools to set up a proper REST API. We might also set up HATEOAS for our API.
For the API we will need a serializer, which will convert the data objects into json/xml or vice versa. Next to that we need a validation library, that will validate the data and also send error messages so we can use them. E.g. think about Symfony Forms and Validation.

## Internal API
A storage Interface should provide a description of all methods required to create, store, retrieve and remove items internally. This way it's possible to more easily change to another storage engine or implement another library.

### Suggestion 1
The storage interface will support versioning, multilingual data, caching and possible indexing. This is done through a set of interfaces which should be implemented for each underlying technology. 

### Suggestion 2
We don't use a flexible way to implement our storage technology and create a sort of super config, similar to Superdesk, in which Models can be described. Their properties, api visibility, validate, serialization and indexing settings are all defined there,

## Article Documents
There are three article documents which together create a final article, namely:
- Article
- Article Metadata
- Article Item

### Article Document
The Article Document contains the main information of an Article.

Fields:
- Id
- Title
- Locale
- Creation date
- Publication date
- Last modified date

### Article Metadata Document
The Article Metadata Document contains all data which should be stored but is not defined in the Article Document.

Fields:
- Article Id
- Locale
- Metadata key
- Metatdata value

### Article Item Document
The Article Item Document forms a link between the created Articles and the ingested Items from Superdesk, which will be the source for most of the article content. If an article is based on a single Item it will have a reference to that, else it will have a reference to the Package (group of items) it is based on.

Fields:
- Article Id
- Item / Package Id

### Item Document
The Item Document is a one on one representation of the items returned by the Content API.

Fields:
- Id
- Name
- Description
- Body
-  + Metadata fields

### Package Document
The Item Document is a one on one representation of the items returned by the Content API.

Fields:
- Id
- Name
- Description
- Items


src/SWP/ContentBundle/Storage/StorageException/StorageDeleteException
```diff
@@ -0,0 +1,24 @@
+<?php
```
ahilles107: missing file extension

src/SWP/ContentBundle/Storage/StorageException/StorageSaveException
```diff
@@ -0,0 +1,24 @@
+<?php
```
ahilles107: missing file extension

thnkloud9:
I would suggest not putting the verb in the api URLs:

GET	        /api/storage/get/	        articles/	article/{id}
POST	/api/storage/create/	        articles/	-
PUT	        /api/storage/update/  	-	article/{id}
DELETE	/api/storage/delete/	        -	article/{id}

to

GET	        /api/storage/	        articles/	article/{id}
POST	/api/storage/	        articles/	-
PUT	        /api/storage/             -	article/{id}
DELETE	/api/storage/            	 -	article/{id}

and possibly even drop the "storage" part as well?  is it needed?

takeit:
well, actually it should be similar as close as possible to superdesk api concept

petrjasek:
agree with @thnkloud9 those verbs are methods, so there should be no need to have it in url

m038:
Updated.

takeit:
apart from that, according to our coversation, we need a way to choose between other storage layers (mongodb, couchdb etc.), bundle configuration should allow for that. 

**So to wrap it up all together I see it like that from the technical point of view:**
- we should create service/manager to manage articles - `ArticleManager` (deleting, saving, finding etc) which can implement `StorageInterface` that was already created. In construct we should be able to pass model class, driver specific manager object (responsible for persisting, flushing data etc) of type `Doctrine\Common\Persistence\ObjectManager` (each library (mongodb, couchdb) implements this interface as it's part of the doctrine library) from which we can get the repository of passed document model.

- based on the provided database driver in the config:

```yaml
swp_storage:
    db_driver: mongodb # couchdb etc
```

we can easly use MongoDB, CouchDB (etc) ManagerRegistry (which extends DoctrineBridge's [ManagerRegistry](https://github.com/symfony/DoctrineBridge/blob/master/ManagerRegistry.php)) class as a factory, to create MongoDB, CouchDB etc. manager object accessible as a service

- based on loaded driver, we can load specific configuration yaml file with `ArticleManager` specific setup for each driver. This setup will define ArticleManager as a service and as a argument will be accepting model class and defined driver manager object (for MongoDB it is `Doctrine\ODM\MongoDB\DocumentManager`, for CouchDB: `Doctrine\ODM\CouchDB\DocumentManager` etc).

This is just a general overview of the idea. Let me know if it makes sense.

ahilles107:
@takeit - all sounds good, but for now as we use PHPCR - only mysql/postgresql will work for us. So i don't know if we don't go to deep here, and instead switching db_driver (mongo, mysql, postgres) we should allow switching adapters - like: phpcr, custom_storage_1, custom_storage_2 etc.

m038:
I agree, but i have some idea's worked out and made some interfaces, i will push them later and we can discuss if that's flexible enough for now.

hlmnrmr:
In regards to the __API Endpoints__, i think it should be simplified as in:

Retrieving a list of articles:
```
GET /api/articles -> provides a default behavior, e.g. returns the latest 10 published articles
GET /api/articles?<query> -> custom queries
```

Retrieving an individual article:
```
GET /api/articles/<item id>
```

Creating an article:
```
POST /api/articles + payload
```

Removing an article:
```
DELETE /api/articles/<item id>
```

Update an existing article:
```
PUT /api/articles/<item id> + payload
```

No need to use plurality and singularity when dealing with the same resource.

m038:
Yeah you are correct. Take a look here, it's my routing configuration for the API https://github.com/m038/web-renderer/blob/SWP-69/src/SWP/ContentBundle/Resources/config/routing.yml and the is the controller skeleton including documentation https://github.com/m038/web-renderer/blob/SWP-69/src/SWP/ContentBundle/Controller/ApiArticleController.php

ahilles107:
Close in favour of #76

PR #29, opened 2015-09-23
Collection pages

ahilles107:


m038:
:+1:

src/SWP/ContentBundle/Tests/Twig/Extension/ContentExtensionTest.php
```diff
@@ -0,0 +1,104 @@
...
+        $this->assertTrue($generateUrlFor, 'Should find "generateUrlFor*" to be a valid function');
+    }
+
+    public function testGetName()
+    {
+        $contentExtension = new ContentExtension(
+            $this->getContainer()->get('doctrine.orm.entity_manager'),
+            $this->getContainer()->get('router')
+        );
+        $this->assertEquals('swp_content', $contentExtension->getName(), 'Should have name "gitHubGist"');
```
takeit: `'Should have name "gitHubGist"'` ? should be be `"swp_content"` not `"gitHubGist"`

src/SWP/FixturesBundle/Resources/themes/theme_1/index.html.twig
```diff
@@ -7,4 +7,11 @@
     <a href="{{ path('swp_page_about_us') }}">About Us</a>
     <a href="{{ path('swp_page_features') }}">Features</a>
     <a href="{{ path('swp_page_get_involved') }}">Get Involved</a>
+
+    <h2>News</h2>
+    <ul>
+    {% gimmelist article from articles with {'pageName': 'News'} %}
+        <li><a href="{{ generateUrlForArticle(article) }}">{{ article.title }} </a></li>
```
takeit: this function should be shorter imho, like `urlFor($article)`, `gimmeUrl($article)` etc., you already check the instance of passed param in that Extension's function. You pass instance of Article as a param, so you already know what you want to get, typing `..Article(article)` makes no sense to me.

src/SWP/ContentBundle/Twig/Extension/ContentExtension.php
```diff
@@ -0,0 +1,75 @@
...
+    {
+        return 'swp_content';
+    }
+
+    public function generateUrlFor($name, $object)
+    {
+        if (is_object($object) && method_exists($object, 'getValues')) {
+            if ($name == 'Article' && $object->getValues() instanceof Article) {
+                $pageArticle = $this->em->getRepository('SWP\ContentBundle\Model\PageContent')
+                    ->getForContentPath($object->getValues()->getId())
```
takeit: `getForContentPath` is not clear, it doesnt tell me what you want to get, query or some value for the contentPath in the form of a query object  etc. that should be `getContentPathQuery()` or if we assume that repository will return only objects of type `Query,` we could simply name it `getContentPath()`

takeit:
PR needs to be also rebased with master, apart from my other comments it's :+1:

src/SWP/FixturesBundle/Resources/themes/theme_1/index.html.twig
ahilles107: :+1: - will change that.

src/SWP/ContentBundle/Twig/Extension/ContentExtension.php
ahilles107: Repository by default should return Query (imo). And agree name is wrong, how about `getByContentPath` ?

takeit: :+1:

takeit:
Scrutinizer reports: https://scrutinizer-ci.com/g/superdesk/web-renderer/inspections/1524cbb1-2a4c-4b88-8c0f-86b4783d6156/issues/files/src/SWP/ContentBundle/Twig/Extension/ContentExtension.php?status=new&orderField=path&order=asc

PR #42, opened 2015-12-09
add composer exclude-from-classmap

takeit:
excludes not needed tests/specs files from classmap, useful in prod env as we don't need there tests

I would suggest to add this option to every bundle/component we created.


ahilles107:
:+1:

PR #43, opened 2015-12-09
Organizations

takeit:
- [x] - configured gedmo doctrine extensions
- [x] - resolves the tenant based on subdomain
- [x] - added doctrine sql filter to set the tenant id on each query
- [x] - added event subscriber to set the Tenant on prePrersist event

...

TODO:
- [ ] - documentation
- [x] - throw errors where the page is not found
- [x] - configure homepage for each tenant
- [x] - more tests
    - [x] - fix phpunit tests for WebrendererBundle
    - [x] - write tests for multi-tenancy configuration, extension and initializer classes
- [x] - multitenancy for phpcr
    - [x] - generate id based on current tenant
    - [x] - created phpcr base paths repository initializer
    - [x] - created configuration and extension to use multitenancy for phpcr
    - [x] - make sure phpcr documents are always tenant aware 
- [x] - fallback to single tenant mode when no subdomains (`default` tenant is created by default)
- [x] - add fixtures
- [ ] - split this into separate repositories (bundle, component, common component)
- [x] - routes aware of tenant
- [ ] - get the ogranization id from the superdesk (not sure yet how that will be solved)
- [ ] - simple factory to create tenants

**REST API** (API setup required form PR in Pawel's PR)

- [ ] - [GET] - Get organizations `/api/organizations`
- [ ] - [GET] - Get single organization `/api/organizations/{id}`
- [ ] - [POST] - Create a new organization `/api/organizations`
- [ ] - [PATCH] - Update organization `/api/organizations`
- [ ] - [DELETE] - Delete organization `/api/organizations/{id}`


app/config/config.yml
```diff
@@ -130,3 +151,14 @@ swp_updater:
 
 swp_bridge:
     base_uri: 'http://superdesk.dev:5050'
+
+stof_doctrine_extensions:
+    default_locale: %locale%
+    orm:
+        default:
+            tree: true
```
ahilles107: turn off not needed extensions, they make booting slower

app/config/parameters.yml.dist
```diff
@@ -19,3 +19,5 @@ parameters:
     secret:            SuperSecretTokenPleaseChangeIt
     swp_updater.version.class: "SWP\WebRendererBundle\Version\Version"
     active_theme: "theme_1"
+    swp_multi_tenancy.tenant.class: SWP\MultiTenancyBundle\Model\TenantInterface
```
ahilles107: why we need this as parameter?

takeit: I can put that directly into the config, seems more reasonable to put it into the parameters to me tho

src/SWP/MultiTenancyBundle/EventListener/TenantListener.php
```diff
@@ -0,0 +1,77 @@
...
+
+    /**
+     * Enable the tenantable filter on kernel.request.
+     *
+     * @param \Symfony\Component\HttpKernel\Event\GetResponseEvent $event
+     */
+    public function onKernelRequest(GetResponseEvent $event)
+    {
+        if (!$event->isMasterRequest()) {
+            return;
```
ahilles107: We don't care on sub requests? Why?

app/config/parameters.yml.dist
ahilles107: but we will change it per instance? If not then it can go to parameters but inside config.yml section `parameters`

src/SWP/MultiTenancyBundle/Resolver/TenantResolver.php
```diff
@@ -0,0 +1,101 @@
...
+     *     throw $this->createNotFoundException('Subdomain not found!');
+     *
+     * @param string          $message  A message
+     * @param \Exception|null $previous The previous exception
+     *
+     * @return NotFoundHttpException
+     *
+     * @see https://github.com/symfony/framework-bundle/blob/master/Controller/Controller.php#L253
+     */
+    protected function createNotFoundException($message = 'Not Found', \Exception $previous = null)
```
ahilles107: if we use it once in this class then we don't need function for one line of code ;)

src/SWP/MultiTenancyBundle/EventListener/TenantListener.php
takeit: good point, will change that

takeit:
closed in favour of https://github.com/superdesk/web-renderer/pull/46

PR #54, opened 2016-02-18
[wip] sylius themes

ahilles107:
replace liip themes with sylius themes bundle, adjust tests and documentation, update composer

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | yes
| Deprecations? | yes
| Tests pass?   | yes
| License       | https://www.superdesk.org/license


app/themes/DefaultTheme/composer.json
```diff
@@ -0,0 +1,8 @@
+{
+    "name": "swp/default-theme",
```
takeit: this should be configured like: 

```
{
    "name": "swp/default-theme",
    "title": "Superdesk Web Publisher default theme"
}
```

we dont use composer to install theme so "extra" property is not needed.

ahilles107: yep

PR #59, opened 2016-02-27
Update docker-compose.yml

actionless:
| Q             | A
| ------------- | ---
| Bug fix?      | [yes]

https://forum.sourcefabric.org/discussion/17892/superdesk-web-publisher-error

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,7 +1,7 @@
 php:
     build: docker/php-fpm
     expose:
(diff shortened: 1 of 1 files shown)


ahilles107:
thanks! :+1:

actionless:
oops, a guy from the superdesk support forum told what it was not enough to make it work with a newer docker-compose version, i will submit more fixes later on then

PR #60, opened 2016-02-28
upgrade sylius/theme-bundle, upgrade project to Symfony2.8 again

ahilles107:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | no
| Fixed tickets | no
| License       | GPLv3

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -45,10 +45,6 @@
     "repositories": [
         {
             "type": "vcs",
-            "url": "https://github.com/Sylius/SyliusResourceBundle"
-        },
-        {
-            "type": "vcs",
             "url": "https://github.com/FriendsOfSymfony/FOSRestBundle"
         }
     ],
@@ -83,8 +79,7 @@
         "stof/doctrine-extensions-bundle": "^1.2",
         "symfony-cmf/routing-bundle": "^1.3",
         "doctrine/doctrine-cache-bundle": "1.0.*",
-        "sylius/theme-bundle": "0.17.x-dev",
-        "sylius/resource-bundle": "0.17.x-dev as 0.17"
+        "sylius/theme-bundle": "0.17.x-dev"
     },
     "require-dev": {
         "sensio/generator-bundle": "~2.3",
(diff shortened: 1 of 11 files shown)


composer.json
```diff
@@ -42,9 +42,19 @@
             "**/Tests/", "**/spec/"
         ]
     },
+    "repositories": [
```
takeit: this should be removed

src/SWP/Bundle/MultiTenancyBundle/EventListener/TenantableListener.php
```diff
@@ -53,8 +53,8 @@ public function __construct(EntityManagerInterface $entityManager, TenantContext
     public function onKernelRequest()
     {
         $tenant = $this->tenantContext->getTenant();
-        $tenantId = $tenant->getId();
-        if ($tenantId) {
+        if ($tenant && $tenant->getId()) {
```
takeit: :+1:

composer.json
```diff
@@ -61,7 +71,7 @@
         "emanueleminotto/twig-cache-bundle": "~1.0",
         "incenteev/composer-parameter-handler": "~2.0",
         "friendsofsymfony/http-cache-bundle": "~1.0",
-        "friendsofsymfony/rest-bundle": "2.0.x-dev",
+        "friendsofsymfony/rest-bundle": "2.0.x-dev as 1.5",
```
takeit: is this for SyliusResourceBundle?

composer.json
ahilles107: Yep

ahilles107: For now it need to stay. We don't control sylius repos, and without it we can't use aliases.

composer.json
ahilles107: yep, it need to stay like that as long as Sylius will support 2.0.x-dev this same with "sylius/resource-bundle"

PR #61, opened 2016-02-28
Update docker-compose.yml

actionless:
| Q             | A
| ------------- | ---
| Bug fix?      | [yes]

this time i tested it -- works fine now

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,6 +1,6 @@
 php:
     build: docker/php-fpm
-    expose:
+    ports:
         - "9000:9000"
     command: composer
     links:
@@ -15,17 +15,17 @@ nginx:
         - php
     volumes:
         - .:/var/www/webrenderer
-        - logs/webrenderer:/var/www/webrenderer/app/logs
-        - logs/nginx/:/var/log/nginx
+        - ./logs/webrenderer:/var/www/webrenderer/app/logs
+        - ./logs/nginx/:/var/log/nginx
     tty: true
 postgres:
     image: postgres:latest
     ports:
         - "5432:5432"
     volumes:
-        - docker/pgsql/data:/var/lib/postgresql
-        - docker/pgsql/config:/etc/postgresql
-        - logs/pgsql:/var/log/postgresql
+        - ./docker/pgsql/data:/var/lib/postgresql
+        - ./docker/pgsql/config:/etc/postgresql
+        - ./logs/pgsql:/var/log/postgresql
     environment:
         LC_ALL: C.UTF-8
         POSTGRES_DB: symfony


actionless:
why it not merged yet?

takeit:
there are still some issues with the `docker-compose.yml`. You can't run Web Publisher when using docker on OSX. Have you tried to run it or you just checked that the config is valid when doing `docker-compose build` ? I was working on some fixes for this which I will push soon. Anyway, this can be merged as its not really related to the main issue.

takeit:
thanks @actionless !

actionless:
you mean there are unresolved things for os x or in general?

takeit:
In general, I didnt test it on, for example, Ubuntu. On osx it definitely doesn't work, even if the configuration seems to be fine, the app cant be properly set up.

actionless:
how can it be verified what it's functional?

takeit:
Have you tried to access the Web Publisher from the IP of your docker machine in your browser, if it works ? Because the Web Publisher vendors are not being installed at all and that might be related to the volumes not working properly on OSX.

actionless:
do i need to provide any special configuration to point to the superdesk instance?

i will try to run it tomorrow locally and do the necessary changes if needed

but even right now i am pretty concerned about the current implementation because it's mounting code to the container instead of bundling it that is not ok by me


also i would suggest to plug webpublisher project to our current Bamboo CI workflow used for other superdesk-based projects (that probably something to discuss with @hlmnrmr )

actionless:
there are several issue atm:
1) that container hanging with `tty: true`
2) some weird problem when mount path is just "."
3) hardcoded nginx vhost
4) as i mentioned above i don't see a point of having two container for php-fpm and nginx and using mounted volume, those containers could be split into one
5) bundle application itself inside the container, composer dependencies are required part of the app so composer install should happen on container build, not run
6) sequence of commands in the Dockerfile is not cache-friendly (but it's a minor issue)

takeit:
@actionless would have some time to take a look into it and try to fix it?

> 5) bundle application itself inside the container, composer dependencies are required part of the app so composer install should happen on container build, not run

But how we would then access files in container? Having volumes allows us to have files locally and modify them very easly while running it on container. (or I don't understand something here)

actionless:
> would have some time to take a look into it and try to fix it?
i can but it'll be one-time fix, i recommend including webpublisher into superdesk bamboo workflow

> But how we would then access files in container? Having volumes allows us to have files locally and modify them very easly while running it on container in a very simple way. 
in superdesk we have a separate docker-compose config for that specific case (which shares the same base named `common.yml` as normal docker-compose config):
https://github.com/superdesk/superdesk/blob/master/docker/docker-compose-dev.yml

Issue #55, opened 2016-02-22
[RFC] Tenant aware themes


takeit:
**Introduction**
- Load themes for current tenant. The repository should not collect all available themes belonging to other tenants - this will be the performance improvement when loading themes for the first time.

**Proposal**

The Sylius Theme Bundle has been refactored a lot, thus the below proposal is not valid anymore. See [this RFC issue](https://github.com/Sylius/Sylius/issues/4280) for more details:

- ~~disable ThemeRepository compiler pass in ThemeBundle~~
- ~~When creating a new tenant (e.g. `tenant1`) it should also  create a new directory in `app/themes/tenant1` where themes will be kept for given tenant. (the dir should be removed when tenant will be deleted)~~
- ~~In `ActiveThemeListener` Event listener (which will be created on our side) there should be `TenantContext` injected to get current tenant. Based on tenant's subdomain we will search in `/app/themes/tenant1` dir for  its available themes (worth to take a look into `FilesystemConfigurationProvider` from ThemeBundle)~~

~~Here we can extend Sylius `RecursiveFileLocator` where we will inject `TenantContext` and prefix the `$paths` array (https://github.com/Sylius/SyliusThemeBundle/blob/master/Locator/RecursiveFileLocator.php#L40) with current tenant, so it will read themes from `app/themes/tenant1` in that case.~~

- ~~found themes should be then added to Sylius `ThemeRepository`~~
- ~~then in the `ThemeRepository` it should find the theme used by the current tenant and set it to active~~

**Drawbacks**

- ~~we should make it configurable to disable/enable ThemeRepositoryPass in Sylius ThemeBundle~~
- ~~remove 'final' from some classes to allow override in Sylius ThemeBundle~~
- ~~make sure the cache should not be cleared when uploading changing themes - ThemeBundle should take care about this? - research it.~~
- ~~dump assets for all themes, not only for current tenant~~

**References**
- [Sylius ThemeBundle](https://github.com/Sylius/SyliusThemeBundle)

ahilles107:
:+1:

PR #65, opened 2016-05-16
[WebRendererBundle, MultiTenancy] Load theme for current tenant

takeit:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | yes
| Deprecations? | no
| Tests pass?   | no
| Fixed tickets | #55, [#SWP-126](https://dev.sourcefabric.org/browse/SWP-126), [#SWP-89](https://dev.sourcefabric.org/browse/SWP-89)
| License       | GPLv3

- [x] - Removed `ActiveThemeListener`, created `TenantAwareThemeContext` instead and enabled it in config so theme will be set before container initialization.

```
sylius_theme:
    context: "swp_webrenderer.theme.context.tenant_aware"
...
```

- [x] - allow to set theme when creating a new tenant via console command [MultiTenancyBundle]
- [x] - load themes only for current tenant [WebRendererBundle]
    - [x] added custom configuration source
    - [x] added custom theme factory and theme model
- [x] - SWP-89 - Create simple factory to create new tenants
- [x] - updated dependencies
- [x] - added `theme_command_test` to `FixturesBundle`, and used it in command setup tests instead of DefaultTheme (demo theme) to not mix things up.
- [x] - [MultiTenancy] SWP-120 - Create Twig global to get current tenant in twig templates (`organization` variable)
- [x] - [WebRenderer] added twig extension - `theme` global variable

TODO:
- [x] - update fixtures
- [x] - update documentation
- [x] - fix tests
- [x] - fix tests on php7
- [x] - phpunit tests for dependency injection classes [MultiTenancyBundle]


src/SWP/Bundle/FixturesBundle/Command/ThemeSetupCommand.php
```diff
@@ -82,14 +84,17 @@ protected function execute(InputInterface $input, OutputInterface $output)
         $helper = $this->getHelper('question');
         $force = true === $input->getOption('force');
 
-        if (!$name) {
+        if (null === $name) {
```
ahilles107: why not `is_null()` ?

src/SWP/Bundle/TemplateEngineBundle/Resources/config/doctrine/ContainerData.orm.yml
```diff
@@ -1,6 +1,6 @@
 SWP\Bundle\TemplateEngineBundle\Model\ContainerData:
     type: entity
-    table: null
+    table: swp_container_data
```
ahilles107: Why you decided to add prefix to table names?

src/SWP/Bundle/WebRendererBundle/Controller/DefaultController.php
```diff
@@ -30,15 +31,17 @@ public function indexAction()
...
+        if (null === $homepage) {
             throw $this->createNotFoundException('No homepage configured!');
         }
 
         $tenantContext = $this->get('swp_multi_tenancy.tenant_context');
+        $themeContext = $this->get('swp_webrenderer.theme.context.tenant_aware');
 
         return $this->render('index.html.twig', [
-            'tenant' => $tenantContext->getTenant(),
+            'tenant' => $tenantContext->getTenant(), // TODO remove, or create twig function
```
ahilles107: I would add it as twig global.

src/SWP/Bundle/WebRendererBundle/Resources/config/services.yml
```diff
@@ -1,16 +1,10 @@
-# override default cmf dynamic router with our implementation (with support for Meta Article)
 parameters:
+    # override default cmf dynamic router with our implementation (with support for Meta Article)
```
ahilles107: Meta Classes (there will be more than Article)

src/SWP/Bundle/WebRendererBundle/Resources/config/services.yml
```diff
@@ -26,3 +20,10 @@ services:
             - "@filesystem"
             - "@kernel"
             - "@swp_webrenderer.detection.device_detection"
+
+    swp_webrenderer.theme.context.tenant_aware:
+        class: SWP\Bundle\WebRendererBundle\Theme\TenantAwareThemeContext
+        # public: false
```
ahilles107: remove commented code

src/SWP/Bundle/WebRendererBundle/Theme/Configuration/TenantableConfigurationProvider.php
```diff
@@ -0,0 +1,81 @@
...
+            $configs = array_map(
+                [$this->loader, 'load'],
+                $this->fileLocator->locateFilesNamed($this->configurationFilename)
+            );
+
+            return array_map(
+                [$this->themeHelper, 'process'],
+                $configs
+            );
+        } catch (\InvalidArgumentException $exception) {
```
ahilles107: why you catch it and throw it again? it's not obsolete?

src/SWP/Component/MultiTenancy/Factory/TenantFactory.php
```diff
@@ -0,0 +1,43 @@
...
+     */
+    public function __construct($className)
+    {
+        $this->className = $className;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function create()
```
ahilles107: as it can't be chained then in factory it's better to make it static. Then instead keeping it in container and fetching if from it every time we could just do:
```
use SWP\Component\MultiTenancy\Model\Tenant;
$tenant = TenantFactory::create(Tenant::class);
```
And as class name is only stored in that service thing then i don't see why it should be in container.

src/SWP/Bundle/TemplateEngineBundle/Resources/config/doctrine/ContainerData.orm.yml
takeit: To easily locate the SWP tables when needed. Note that some vendors can create tables too, prefixes makes it more clear to see what is part of the SWP and what's not when it comes to the database. It's a good practice. There is also a possibility to create a custom listener that will apply the prefix automatically so we could make it also configurable if needed, but I don't think it's a blocker :)

src/SWP/Bundle/WebRendererBundle/Resources/config/services.yml
takeit: you wrote it :) I just moved, I will correct it :P

src/SWP/Bundle/WebRendererBundle/Theme/Configuration/TenantableConfigurationProvider.php
takeit: good catch, its because I wanted to silence it and moved the logic to custom theme context, but I think here its fine that it throws it.

src/SWP/Component/MultiTenancy/Factory/TenantFactory.php
takeit: but then it means, in the code you will need to use `$tenant = TenantFactory::create(Tenant::class);` everywhere where you will want to instantiate a new tenant object. If for some reason you will change that class, its method etc. you will need to change it everywhere you used it. Why not use Symfony container? also static = evil.

PR #76, opened 2016-06-02
SWP-60 Design and implement storage

takeit:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | yes
| Deprecations? | yes
| Tests pass?   | yes
| Fixed tickets | [SWP-60](https://dev.sourcefabric.org/browse/SWP-60), [SWP-59](https://dev.sourcefabric.org/browse/SWP-59), [SWP-29](https://dev.sourcefabric.org/browse/SWP-29), [SWP-141](https://dev.sourcefabric.org/browse/SWP-141), [SWP-142](https://dev.sourcefabric.org/browse/SWP-142), [SWP-143](https://dev.sourcefabric.org/browse/SWP-143), [SWP-145](https://dev.sourcefabric.org/browse/SWP-145), [SWP-184](https://dev.sourcefabric.org/browse/SWP-184)
| License       | AGPLv3

**BC breaks:**

- re-create the database

**Depracations:** 
- `content` property has been renamed to `body` in Article model - I didn't add the depracation note because we dont release yet

**Changes:**

- ContentBundle
    - [x] - make the bundle storage agnostic (Doctrine ORM, Doctrine PHPCR, we can add later Doctrine ODM support (ie. MongoDb etc)) - load mapping files depending on enabled persistence backend
    - [x] - add phpcr mapping
    - [x] - add content manager so it can work independently from the persistance backend (ORM, PHPCR etc.) - to save, fetch articles etc

- BridgeBundle
    - [x] - add Bridge component
    - [x] - add Package and Item basic models
    - [x] - implement validators to validate incoming payload[SWP-143](https://dev.sourcefabric.org/browse/SWP-143) 
    - [x] - implement transformers:
         -  [x] - json payload <=> Package/Item
         -  [x] - Package/Item <=> Article

- Common Component
    - [x] add timestampable, translatable interface and trait

- Storage Bundle
    - [x] - register repositories as a services independantly from the selected persistance backend
    - [x] - storage agnostic repositories
    - [x] - drivers implementation - (PHPCR atm)
    - [x] - create storage agnostic factory to instantiate objects independently from the selected persistence backend

- Storage Component
    - [x] - added generic interfaces, classes to build storage agnostic layer
    - [x] - load doctrine mapping files automatically

**TODO:**
 - [x] - add bridge component to git subsplit script
 - [x] - add storage component and storage bundle to git subsplit script
 - [x] - create route events
 - [x] - create article events
[...]

PR #78, opened 2016-06-09
Thnkloud9analytics

djbrd-sourcefabric:
Simply starts a stopwatch on the kernel.request event, ends it on the kernel.response event, and saves the result with the name of the route, and the uri, in the db.
I suggest that we configure which routes we want to show with a configuration file, which consists of a list of routes, each with a number representing the percentage of requests to that route we wish to measure. I would not maintain a count of how many times a route has been requested, but just compare the resutlt of rand(0, 100) with the given percentage to decide whether to measure the duration of the current request.


.gitignore
```diff
@@ -19,3 +19,4 @@
 /web/config.php
 logs/
 build/
+.idea/
```
takeit: this is not bundle related, but phpstorm hidden files, you can exclude it in your local gitignore instead of here

app/Resources/views/index.html.twig
```diff
@@ -0,0 +1,6 @@
+{% extends 'base.html.twig' %}
```
takeit: should be remove, we use themes now so this file is not needed

src/SWP/Bundle/AnalyticsBundle/DependencyInjection/SWPAnalyticsExtension.php
```diff
@@ -0,0 +1,41 @@
+<?php
+
+/**
+ * This file is part of the Superdesk Web Publisher Web Renderer Bundle.
+ *
+ * Copyright 2015 Sourcefabric z.u. and contributors.
```
takeit: 2016 and `z.ú`

src/SWP/Bundle/AnalyticsBundle/EventListener/KernelListener.php
```diff
@@ -0,0 +1,97 @@
+<?php
+
+/**
+ * This file is part of the Superdesk Web Publisher Template Engine Bundle.
+ *
+ * Copyright 2015 Sourcefabric z.u. and contributors.
```
takeit: 2016 and z.ú

src/SWP/Bundle/AnalyticsBundle/EventListener/KernelListener.php
```diff
@@ -0,0 +1,97 @@
...
+     * @var Stopwatch
+     */
+    protected $stopwatch;
+
+    /**
+     * @var StopwatchEvent
+     */
+    protected $stopwatchEvent;
+
+
```
takeit: remove extra space

src/SWP/Bundle/AnalyticsBundle/EventListener/KernelListener.php
```diff
@@ -0,0 +1,97 @@
...
+    /**
+     * @var RequestMetricRepository
+     */
+    protected $requestMetricRepository;
+
+    public function __construct()
+    {
+        $this->stopwatch = null;
```
takeit: by default the property's value is `null` so no need to add construct to declare it, or easier would be to do `protected $stopwatchEvent = null;`

src/SWP/Bundle/AnalyticsBundle/EventListener/KernelListener.php
```diff
@@ -0,0 +1,97 @@
...
+
+    public function onKernelRequest(GetResponseEvent $event)
+    {
+        if (!$event->isMasterRequest()) {
+            return;
+        }
+
+        $route = $event->getRequest()->get('_route');
+
+        // Todo: check for configured routes
```
takeit: todo should be done? :)

src/SWP/Bundle/AnalyticsBundle/EventListener/KernelListener.php
```diff
@@ -0,0 +1,97 @@
...
+    {
+        if (is_null($this->stopwatch)) {
+            return;
+        }
+
+        $requestUri = $event->getRequest()->getUri();
+
+        // find out if this is a profiler request, if not don't analyze it
+        try {
+            $this->stopwatchEvent = $this->stopwatch->stop($requestUri);
```
takeit: how does that finds if this is a profiler request? you just catch the exception

src/SWP/Bundle/AnalyticsBundle/Model/RequestMetric.php
```diff
@@ -0,0 +1,136 @@
...
+ * For the full copyright and license information, please see the
+ * AUTHORS and LICENSE files distributed with this source code.
+ *
+ * @copyright 2015 Sourcefabric z.ú.
+ * @license http://www.superdesk.org/license
+ */
+namespace SWP\Bundle\AnalyticsBundle\Model;
+
+/**
+ * RequestMetric -.
```
takeit: this comment can be removed if its the same a class name, it doesnt tell much anyway

src/SWP/Bundle/AnalyticsBundle/Model/RequestMetric.php
```diff
@@ -0,0 +1,136 @@
...
+     *
+     * @param int $duration
+     *
+     * @return RequestMetric
+     */
+    public function setDuration($duration)
+    {
+        $this->duration = $duration;
+
+        return $this;
```
takeit: do you really need a fluent interface here? If you don't use it anywhere in the bundle then just `$this->duration = $duration;` is enough, pls check also in other places

src/SWP/Bundle/AnalyticsBundle/Repository/RequestMetricRepository.php
```diff
@@ -0,0 +1,28 @@
...
+namespace SWP\Bundle\AnalyticsBundle\Repository;
+
+use SWP\Bundle\AnalyticsBundle\Model\RequestMetric;
+
+class RequestMetricRepository extends \Doctrine\ORM\EntityRepository
+{
+    /**
+     * @param RequestMetric $requestMetric Persists object and flushes the entity manager
+     */
+    public function save(RequestMetric $requestMetric)
```
takeit: I would name that method `add` as it's better reflecting the operation you want to do, `save` would mean you want to update the resource and the repository should not be capable of that.

src/SWP/Bundle/AnalyticsBundle/Resources/config/services.yml
```diff
@@ -0,0 +1,14 @@
+services:
+    swp.analytics.requestmetric_repository:
```
takeit: in https://github.com/superdesk/web-publisher/pull/76 I created some logic responsible for automatically registering repositories as a services independently from the selected persistence backend, but this can be merged like that for now if the mentioned PR wont be merged before this one

src/SWP/Bundle/WebRendererBundle/Controller/DefaultController.php
```diff
@@ -34,8 +36,13 @@ public function indexAction()
             throw $this->createNotFoundException('No homepage configured!');
         }
 
-        return $this->render('index.html.twig', [
+        $tenantContext = $this->get('swp_multi_tenancy.tenant_context');
+
+        $response = $this->render('index.html.twig', [
+            'tenant' => $tenantContext->getTenant(),
```
takeit: tenant context should be removed from that controller as it has been already removed before in the project - it's not needed, we have twig global to get the tenant from twig templates

src/SWP/Bundle/AnalyticsBundle/EventListener/KernelListener.php
djbrd-sourcefabric: $this->stopWatch will be null if it's a profiler request

src/SWP/Bundle/AnalyticsBundle/Model/RequestMetric.php
djbrd-sourcefabric: This is just standard practice in any Symfony/Doctrine project I've worked on

src/SWP/Bundle/AnalyticsBundle/EventListener/KernelListener.php
takeit: then what about checking if its null and then return false? instead of catching the exception?

src/SWP/Bundle/AnalyticsBundle/EventListener/KernelListener.php
djbrd-sourcefabric: Hasn't been decided yet how we will check for which routes should be measured. I suggested a configuration file. Pawel wants to do it with a query parameter, so there would be no configuration, although I don't know how such a parameter would be added.

src/SWP/Bundle/AnalyticsBundle/EventListener/KernelListener.php
djbrd-sourcefabric: I do check at the top of the function. Is a return value expected?

takeit: ah ok then it should be fine

src/SWP/Bundle/AnalyticsBundle/EventListener/KernelListener.php
```diff
@@ -0,0 +1,91 @@
...
+    {
+        if (!$event->isMasterRequest()) {
+            return;
+        }
+
+        $route = $event->getRequest()->get('_route');
+
+        // Todo: check for configured routes
+        // Ignore web debug toolbar
+        if (!is_null($route) && $route !== '_wdt') {
```
takeit: Should be:

```php
if (null !== $route && '_wdt' !== $route)
```
See our [coding standards](http://superdesk-web-publisher.readthedocs.io/en/latest/contributing/code/standards.html#structure)

src/SWP/Bundle/AnalyticsBundle/EventListener/KernelListener.php
```diff
@@ -0,0 +1,91 @@
...
+            $this->stopwatch = new Stopwatch();
+            $this->stopwatch->start($event->getRequest()->getUri());
+        }
+    }
+
+
+    public function onKernelResponse(FilterResponseEvent $event)
+    {
+        if (is_null($this->stopwatch)) {
```
takeit:
```php
if (null === $this->stopwatch)
```

src/SWP/Bundle/AnalyticsBundle/EventListener/KernelListener.php
```diff
@@ -0,0 +1,91 @@
...
+        } catch (\LogicException $e) {
+            // do nothing, we shouldn't be analyzing this controller
+            return false;
+        }
+
+        $duration = $this->stopwatchEvent->getDuration();
+        $route = $event->getRequest()->get('_route');
+
+        // Route is null here in test environment
+        if (!is_null($route)) {
```
takeit:
```php
if (null !== $route)
```

src/SWP/Bundle/AnalyticsBundle/Repository/RequestMetricRepository.php
```diff
@@ -0,0 +1,28 @@
...
+namespace SWP\Bundle\AnalyticsBundle\Repository;
+
+use SWP\Bundle\AnalyticsBundle\Model\RequestMetric;
+
+class RequestMetricRepository extends \Doctrine\ORM\EntityRepository
+{
+    /**
+     * @param RequestMetric $requestMetric Persists object and flushes the entity manager
+     */
+    public function persistAndFlush(RequestMetric $requestMetric)
```
ahilles107: Why method like that? Persisting and flushing should happen in service (or if there is no service yet then directly in controller). Repository should return Queries or Objects, never saving/persisting/modifying objects.

djbrd-sourcefabric: To avoid injecting the entity manager into the service, as suggested here:
http://php-and-symfony.matthiasnoback.nl/2014/05/inject-a-repository-instead-of-an-entity-manager/

takeit: repository actually can persist, remove data, but it should not **update** the data.

ahilles107: Ok, sounds good after all.

djbrd-sourcefabric: It doesn't update the data. It persists the data, and flushes the entity
manager. The guy who writes that blog also writes books on Symfony, so he
probably hasn't suggested anything which isn't OK. Injecting the entity is
generally a bad idea, as it can easily lead to circular references in my
experience.

src/SWP/Bundle/AnalyticsBundle/EventListener/KernelListener.php
```diff
@@ -0,0 +1,91 @@
...
+            $this->stopwatchEvent = $this->stopwatch->stop($requestUri);
+        } catch (\LogicException $e) {
+            // do nothing, we shouldn't be analyzing this controller
+            return false;
+        }
+
+        $duration = $this->stopwatchEvent->getDuration();
+        $route = $event->getRequest()->get('_route');
+
+        // Route is null here in test environment
```
ahilles107: Why it's like that? If there can be a request without route then better would be to move that check also to top. There is no point to play with other things when at end you can't save it.

src/SWP/Bundle/AnalyticsBundle/EventListener/KernelListener.php
```diff
@@ -0,0 +1,91 @@
...
+    {
+        if (!$event->isMasterRequest()) {
+            return;
+        }
+
+        $route = $event->getRequest()->get('_route');
+
+        // Todo: check for configured routes
+        // Ignore web debug toolbar
+        if (null !== $route && '_wdt' !== $route) {
```
ahilles107: why we wan't to ignore _wdt routes? It's separate requests, so it shouldn't mix anything. And on production it will be not the case. Anyway - we wan't to mark requests/controllers for measuring so those shouldn't be touched also. I think that this check is not needed.

src/SWP/Bundle/AnalyticsBundle/EventListener/KernelListener.php
djbrd-sourcefabric: This is only a problem in the test environment, as far as I can tell. The
parameter is then set in the kernel.request event, but not in the
kernel.response event.

src/SWP/Bundle/AnalyticsBundle/EventListener/KernelListener.php
```diff
@@ -0,0 +1,91 @@
...
+namespace SWP\Bundle\AnalyticsBundle\EventListener;
+
+use SWP\Bundle\AnalyticsBundle\Model\RequestMetric;
+use SWP\Bundle\AnalyticsBundle\Repository\RequestMetricRepository;
+use Symfony\Component\HttpKernel\Event\FilterResponseEvent;
+use Symfony\Component\HttpKernel\Event\GetResponseEvent;
+use Symfony\Component\Stopwatch\Stopwatch;
+use Symfony\Component\Stopwatch\StopwatchEvent;
+
+class KernelListener
```
ahilles107: That name is wrong - It should be `MetricsListener`.

src/SWP/Bundle/AnalyticsBundle/EventListener/KernelListener.php
djbrd-sourcefabric: Well, I just wanted to avoid filling the database with junk in the dev
environment. Anyway, if we are configuring routes to be analysed, this line
will become redundant.

src/SWP/Bundle/AnalyticsBundle/EventListener/KernelListener.php
djbrd-sourcefabric: I assumed we would be adding to the functionality in this bundle. I would
prefer to have just one kernel listener in the AnalyticsBundle. There is
also a chance that there may have to be other classes in the EventListener
folder to handle the metrics, as it's not possible, afaik, to handle
different kinds of events in the same class. So we might have a kernel
listener and a doctrine listener which both handle metrics. So would you
like the class to be renamed MetricsKernelListener, or left as it is until
we have another KernelListener in this bundle?

ahilles107: `as it's not possible, afaik, to handle different kinds of events in the same class` - it is with http://symfony.com/doc/current/cookbook/doctrine/event_listeners_subscribers.html 

Class name should be about functionality not about events, as you should be able to use it also for other events in future (not kernel but custom events).

djbrd-sourcefabric: In that example you have multiple classes listening to the same kind of
events (doctrine events). Can the same classes be used to listen to kernel
events?

ahilles107: All of them are jest events dispatched by event dispatcher. Subscriber/listener doesn't care what event it is (as long as you method don't expect particular even class).

src/SWP/Bundle/AnalyticsBundle/EventListener/KernelListener.php
djbrd-sourcefabric: Well, I just wanted to avoid filling the database with junk in the dev
environment. Anyway, if we are configuring routes to be analysed, this line
will become redundant.

PR #81, opened 2016-06-30
Added JMS Translation Bundle, translation strings to index.html in de…

djbrd-sourcefabric:
| Q             | A
| ------------- | ---
| Bug fix?      | [no]
| New feature?  | [yes]
| BC breaks?    | [no]
| Deprecations? | [no]
| Tests pass?   | [yes]
| Fixed tickets | [SWP-164]
| License       | AGPLv3

…fault theme, and initial xlf file for translation to English and German

The added bundle offers the following functionality: http://jmsyst.com/bundles/JMSTranslationBundle/master/usage
Because of peculiarities in the formatting of the xlf files which it generates, and the verbosity of the cli commands, as we need to use them with Sylius themes, I suggest we consider either doing our own implementation, or extending the bundle. We could also added a twig extension function because the suggested syntax is pretty ugly, I think. I have only so far added translation to index.html.twig, but will change the other files if we decide that this is the solution we want. It does meet our basic requirements, even if we might want to revisit it later to develop one which is more user friendly for long term use.

### DIFF
--- a/app/AppKernel.php
+++ b/app/AppKernel.php
@@ -28,6 +28,7 @@ class AppKernel extends Kernel
             new Stof\DoctrineExtensionsBundle\StofDoctrineExtensionsBundle(),
             new Symfony\Cmf\Bundle\RoutingBundle\CmfRoutingBundle(),
             new Doctrine\Bundle\DoctrineCacheBundle\DoctrineCacheBundle(),
+            new JMS\TranslationBundle\JMSTranslationBundle(),
 
             new SWP\Bundle\MultiTenancyBundle\SWPMultiTenancyBundle(),
             new SWP\Bundle\TemplateEngineBundle\SWPTemplateEngineBundle(),
@@ -35,7 +36,7 @@ class AppKernel extends Kernel
             new SWP\UpdaterBundle\SWPUpdaterBundle(),
             new SWP\Bundle\BridgeBundle\SWPBridgeBundle(),
             new SWP\Bundle\ContentBundle\SWPContentBundle(),
-            new SWP\Bundle\AnalyticsBundle\SWPAnalyticsBundle(),
+            new SWP\Bundle\AnalyticsBundle\SWPAnalyticsBundle()
         );
 
         if (in_array($this->getEnvironment(), ['dev', 'test'])) {
--- a/app/AppKernel.php
+++ b/app/AppKernel.php
@@ -36,7 +36,7 @@ class AppKernel extends Kernel
             new SWP\UpdaterBundle\SWPUpdaterBundle(),
             new SWP\Bundle\BridgeBundle\SWPBridgeBundle(),
             new SWP\Bundle\ContentBundle\SWPContentBundle(),
-            new SWP\Bundle\AnalyticsBundle\SWPAnalyticsBundle()
+            new SWP\Bundle\AnalyticsBundle\SWPAnalyticsBundle(),
         );
 
         if (in_array($this->getEnvironment(), ['dev', 'test'])) {
--- a/app/config/config.yml
+++ b/app/config/config.yml
@@ -9,7 +9,9 @@ parameters:
     locale: en
 framework:
     #esi:             ~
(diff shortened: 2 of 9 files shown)


ahilles107:
Looks ok, but please merge master latest changes back to your branch (this PR).

djbrd-sourcefabric:
Done

On Thu, Jun 30, 2016 at 2:49 PM, Paweł Mikołajczuk <notifications@github.com

ahilles107:
Please add short chapter to our docs for that feature. How to use it by templator and developer (with examples).

djbrd-sourcefabric:
Unfortunately translations in sylius themes do not integrate at all well with debug tools in the Symfony framework. The principle course of this appears to be the way in which the ThemeAwareTranslator class appends the name of the theme to a given locale as a key for storing translations. I assume that this is to avoid conflicts when there is more than one theme. Logging is enabled by default in the translator component. Any string for which there is no translation is logged. Unfortunately, in the logger, there is an assumption that translations will be stored in a library with the locale as a key. So, every translation in a theme is reported as missing. The logger can be turned off, so this is not too serious, but we do lose this functionality. The profiler reports missing translations using a different code path. Again, it reports every translation in a theme as missing.  This is annoying in the development environment. Again, it can be ignored, but we lose this functionality. Finally, the translation:debug console command, which I've found useful in the past, uses hard coded paths and will not find any of the translations in the theme.
So, I suggest we write our own debug:translation command, to be able to debug translations in a theme - to find which are missing, and what are the translations.
Regarding the use of JMSTranslationBundle, the translation:extract command is useful, but does add quite a bit of metadata to XLIFF files which may well be unnecessary, and makes them look very cluttered, and harder to read, I think. Since the command uses code in the Symfony framework in order to extract keys from php and twig files, I wouldn't need more than a day to duplicate this functionality, but without the addition of all the metadata.
The translation:list-resources command in the same bundle is useless for the same reason as the debug commands in the Symfony framework - it uses hard coded paths. It would be easy enough to duplicate this functionality without the hard coded paths.
So. we have a few questions:
Do we want to override the default behaviour in the ThemeAwareTranslator and store messages in a library with the locale as a key, and not the locale + theme_name? This would fix the logging, and the profiler, but might cause other problems - I'm not sure of behaviour with multiple themes.
Or do we want to override the logger to handle messages with the different key?
Or should we just switch off the logger? I vote for this.
Should we also try to switch off the profiling of translations? This isn't necessary but would mean that anyone using the profiler wouldn't have to ignore a lot of warnings.
Or should we try to adapt the profiler?
I strongly suggest that we implement our own debug:translation command. We could reuse the code in functional tests (to check that a translation exists for every key used in a twig or php file).
Given the time, I would like to implement our own translation:extract command.
I think that we can do without the translation:list-resources command.
We could try to implement a standalone bundle for working with translations with sylius themes. Anyone else working with sylius themes will have the same problems, as far as I can tell.
Finally, do we want a solution for translations JavaScript files? I've found a couple of Symfony bundles which provide one, but none from a reputable source. I guess we need to ask whether implementers will want to add strings to a page in JavaScript.
I suggest that we discuss these issues, write tickets, assign priorities, etc.

ahilles107:
@danielhjames please add some formatting to previous comment (lists, paragraphs, headlines etc.) - it's very hard to read it now.

djbrd-sourcefabric:
It would probably be easier for us to talk about this with a screen share.

ahilles107:
> The principle course of this appears to be the way in which the ThemeAwareTranslator class appends the name of the theme to a given locale as a key for storing translations. I assume that this is to avoid conflicts when there is more than one theme.

Every theme have his own translations, and they should be loaded only for current theme (so in one request we don't load translations for every theme).

ahilles107:
If it's possible - switch off the logger and pane in profiller for translations. Ignore translation:list-resources command.

takeit:
is it ready to be merged?

djbrd-sourcefabric:
Yeah

PR #82, opened 2016-07-04
Custom exception thrown and handled when default theme is not installed

djbrd-sourcefabric:
| Q             | A
| ------------- | ---
| Bug fix?      | [yes]
| New feature?  | [no]
| BC breaks?    | [no]
| Deprecations? | [no]
| Tests pass?   | [yes]
| Fixed tickets | [SWP-150]
| License       | AGPLv3

As agreed with Pawel, new functionality is for a custom exception to be thrown if there is no index.html.twig in the default theme. This exception is caught by a kernel listener which displays instructions to install the default theme. Another exception is handled elsewhere, when a theme.json file is not found in the default theme folder, during the bootstrap phase. I am not sure whether this occurs after a clean install. This exception is silently handled. I tried throwing the same custom exception but it was not caught by the kernel listener, probably because it was thrown during the bootstrap phase.


src/SWP/Bundle/WebRendererBundle/Controller/DefaultController.php
```diff
@@ -34,9 +35,13 @@ public function indexAction()
             throw $this->createNotFoundException('No homepage configured!');
         }
 
-        $response = $this->render('index.html.twig', [
-            'page' => $homepage,
-        ]);
+        try {
+            $response = $this->render('index.html.twig', [
```
ahilles107: Nope. This is not what we want. If there is no file then normal exception should be raised (as it was before). We should react earlier (on place where we resolve theme name from tenant and fetch theme - there we should check if theme files exists and if not then raise custom exception).

ahilles107: It should be outside of controller.

src/SWP/Bundle/WebRendererBundle/Theme/Configuration/TenantableConfigurationProvider.php
```diff
@@ -63,14 +63,19 @@ public function __construct(
      */
     public function getConfigurations()
     {
-        $configs = array_map(
-            [$this->loader, 'load'],
-            $this->fileLocator->locateFilesNamed($this->configurationFilename)
-        );
+        // Handle there being no theme.json file
+        try {
```
ahilles107: Not sure why we would need silence errors here.

djbrd-sourcefabric: An invalid argument exception is thrown hair during the kernel boot, with the message that "The file "theme.json" does not exist (searched in the following directories: /home/sourcefabric/Public/web-publisher/app/themes)." This is really where we should throw the NoThemeException, but it can't be handled gracefully at this point. A kernel exception listener will not be called until the kernel is booted.

src/SWP/Bundle/WebRendererBundle/Controller/DefaultController.php
```diff
@@ -37,7 +38,6 @@ public function indexAction()
         $response = $this->render('index.html.twig', [
             'page' => $homepage,
         ]);
-
```
ahilles107: this empty line should stay

src/SWP/Bundle/WebRendererBundle/Theme/Configuration/TenantableConfigurationProvider.php
ahilles107: hmm... @takeit any ideas? I don't like that try..catch and just hiding this error.

src/SWP/Bundle/WebRendererBundle/EventListener/ExceptionListener.php
```diff
@@ -0,0 +1,30 @@
...
+ * @copyright 2016 Sourcefabric z.ú.
+ * @license http://www.superdesk.org/license
+ */
+namespace SWP\Bundle\WebRendererBundle\EventListener;
+
+use SWP\Bundle\WebRendererBundle\Exception\NoThemeException;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\HttpKernel\Event\GetResponseForExceptionEvent;
+
+class ExceptionListener
```
takeit: I would name it `ThemeNotFoundExceptionListener` or `NoThemeExceptionListener` - `ExceptionListener` is too generic imo

src/SWP/Bundle/WebRendererBundle/Exception/NoThemeException.php
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * For the full copyright and license information, please see the
+ * AUTHORS and LICENSE files distributed with this source code.
+ *
+ * @copyright 2016 Sourcefabric z.ú.
+ * @license http://www.superdesk.org/license
+ */
+namespace SWP\Bundle\WebRendererBundle\Exception;
+
+class NoThemeException extends \Exception
```
takeit: what about:

```php
class ThemeNotFoundException extends \RuntimeException
{
    /**
     * {@inheritdoc}
     */
    public function __construct(\Exception $previousException = null)
    {
        parent::__construct('Theme could not be found!', 0, $previousException);
    }
}
```

since you are throwing that exception it would be nice to provide custom message already inside the exception.

src/SWP/Bundle/WebRendererBundle/Theme/Configuration/TenantableConfigurationProvider.php
takeit: I would simply do try catch here ThemeBundle Exception (which is default \InvalidArgumentException) and throw our exception (`ThemeNotFoundException`) instead of ThemeBundle's one. If there are no themes in themes directory it should throw our exception telling that Theme cant be found.

In the `TenantAwareThemeContext` we should throw ThemeNotFoundException (as it is now) and the listener will automatically catch it and display message to install default theme, because it is available at this time already in themes folder and user should see a message just to install it

takeit: I just realized that when throwing exception here it, we will see white page in production env. Silencing it here is the only option I see for now. But the listener should either notify that there are no themes added at all or that user should install default theme. Themes are loaded on kernel boot so to display a message it should pass this code and exception should be thrown later as it is now in this PR

src/SWP/Bundle/WebRendererBundle/EventListener/ExceptionListener.php
djbrd-sourcefabric: I think it's better practice to create a generic exception event listener, in case we want to handle any other exception in this way. We should probably handle more exceptions in this way.

src/SWP/Bundle/WebRendererBundle/Theme/Configuration/TenantableConfigurationProvider.php
djbrd-sourcefabric: The listener? Listeners don't work during the boot phase, I think.

src/SWP/Bundle/WebRendererBundle/EventListener/ExceptionListener.php
ahilles107: But cost (performance) of running more that one listeners like that is lower than time cost of maintaining one big listener for many cases.

takeit: Then in this case a new listener class would be a better solution than putting everything into one class, easier to maintain, test and read.

src/SWP/Bundle/WebRendererBundle/Theme/Configuration/TenantableConfigurationProvider.php
takeit: yes listener, I meant that you need to improve the message in the `ExceptionListener` telling that there are not themes at all or that you need to install demo theme. It will be called later after the container is initializer in the theme context class where you throw an exception.

src/SWP/Bundle/WebRendererBundle/Exception/NoThemeException.php
djbrd-sourcefabric: I don't see the point of adding a message here if we're handling the exception and displaying a page.

takeit: Agree. In case you silence the exception and pass it to the listener it not required

src/SWP/Component/MultiTenancy/spec/Factory/TenantFactorySpec.php
```diff
@@ -22,22 +22,22 @@
...
     }
 
-    function it_implements_tenant_factory_interface()
+    public function it_implements_tenant_factory_interface()
     {
         $this->shouldImplement('SWP\Component\MultiTenancy\Factory\TenantFactoryInterface');
     }
 
-    function it_creates_a_new_empty_tenant()
+    public function it_creates_a_new_empty_tenant()
```
takeit: `public` is not needed in the spec files, it is useless, no need to add it.

djbrd-sourcefabric: This was added by StyleCI automatic fixing

takeit:
is it ready to be merged?

src/SWP/Bundle/WebRendererBundle/EventListener/NoThemeExceptionListener.php
```diff
@@ -0,0 +1,30 @@
+<?php
+/**
+ * This file is part of the Superdesk Web Publisher Web Renderer Bundle.
+ *
+ * Copyright 2016 Sourcefabric z.u. and contributors.
```
takeit: `z.ú.`

src/SWP/Bundle/WebRendererBundle/Exception/NoThemeException.php
```diff
@@ -0,0 +1,18 @@
+<?php
+
+/**
+ * This file is part of the Superdesk Web Publisher Web Renderer Bundle.
+ *
+ * Copyright 2016 Sourcefabric z.u. and contributors.
```
takeit: `z.ú.`

takeit:
@djbrd-sourcefabric please, pull latest changes from master branch

PR #90, opened 2016-07-21
Provide way to set http cache lifetime for routes and expose API for controlling it. 

djbrd-sourcefabric:
| Q             | A
| ------------- | ---
| Bug fix?      | [no]
| New feature?  | [yes]
| BC breaks?    | [no]
| Deprecations? | [no]
| Tests pass?   | [yes]
| Fixed tickets | [SWP-30]
| License       | AGPLv3

Enables user to set how long in seconds a page served by a user defined route will be cached. The s-maxage and max-age are initialised with this value by a listener which listens for a response from the user defined route.


src/SWP/Bundle/ContentBundle/Doctrine/ODM/PHPCR/Route.php
```diff
@@ -58,4 +63,20 @@ public function setType($type)
     {
         $this->type = $type;
     }
+
+    /**
+     * @return int
+     */
+    public function getCacheTimeInSeconds()
```
takeit: I would put this method name in the `RouteInterface`, same for the setter. You would force to implement that method when using models not only for PHPCR

src/SWP/Bundle/ContentBundle/EventListener/HttpCacheHeaderListener.php
```diff
@@ -0,0 +1,41 @@
+<?php
+
+/**
+ * This file is part of the Superdesk Web Publisher Content Bundle.
+ *
+ * Copyright 2015 Sourcefabric z.ú. and contributors.
```
takeit: 2016

src/SWP/Bundle/ContentBundle/Resources/config/services.yml
```diff
@@ -31,6 +31,11 @@ services:
         tags:
             - { name: kernel.event_listener, event: kernel.request, method: onKernelRequest }
 
+    swp_content_bundle.listener.http_cache_header:
+        class: SWP\Bundle\ContentBundle\EventListener\HttpCacheHeaderListener
```
takeit: put `SWP\Bundle\ContentBundle\EventListener\HttpCacheHeaderListener` into the `parameters` section in this file and just put the reference to that parameter here, so it will be easier to override that class in the future, for example by 3rd parties when needed.

```yaml
parameters:
    swp_content_bundle.listener.http_cache_header.class: SWP\Bundle\ContentBundle\EventListener\HttpCacheHeaderListener

swp_content_bundle.listener.http_cache_header:
    class: '%swp_content_bundle.listener.http_cache_header.class%'
```

PR #92, opened 2016-07-27
[Theme][WebRendererBundle] SWP-200 Asset function is not theme-aware

takeit:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | [SWP-200](https://dev.sourcefabric.org/browse/SWP-200)
| License       | AGPLv3

This makes the assets theme-aware. Every asset should be stored under the public directory. See [theme structure documentation](http://superdesk-web-publisher.readthedocs.io/en/latest/themes.html#structure) for more details. The default implementation of SyliusThemeBundle allows only to install assets for bundles. This PR makes it possible to install assets out of the bundle scope.

**How to place and install assets:** (an example)
1. Put the `test.css` asset file in `<theme-name>/public/css/` (create css directory if not exists)
2. Install assets by running command: `php app/console sylius:theme:assets:install`
3. Make use of the asset file in twig template:

```twig
<link rel="stylesheet" href="{{ asset('bundles/css/test.css') }}" />
```

**To see this PR in action run:**

```bash
php app/console theme:setup -f
php app/console sylius:theme:assets:install
```


ahilles107:
how hard would be kicking out `bundles` from `{{ asset('bundles/css/test.css') }}`?

takeit:
It would require to override `PathResolver` class from SyliusThemeBundle to ignore `bundles` prefix from path and make sure all works fine

ahilles107:
If we will leave bundle here now then it will be BC break later for templators. And `bundles` word doesn't make here any sense. Let's remove it.

takeit:
what about renaming `bundles` to `theme` prefix instead? The reason for this is that in the `web` directory you could also put some global assets files (according to Symfony documentation), which you would refer to them using e.g. `{{ asset('css/test.css') }}', '{{ asset('/test2.css') }}` etc. To distinquish those "global" assets from theme assets, changing the prefix to `theme` should be fine and makes it more clear.

```twig
{# loads theme's asset file (e.g. `<theme-name>/public/test.css`) #}
{{ asset('theme/css/test.css') }} 
```

```twig
{# loads asset file directly from `web` dir (`web/test.css`) #}
{{ asset('/css/test.css') }}
```

what do you think?

ahilles107:
Even better!

src/SWP/Bundle/WebRendererBundle/Theme/Asset/AssetsInstaller.php
```diff
@@ -0,0 +1,119 @@
...
+            }
+        }
+
+        return $sources;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function installThemedBundleAssets(ThemeInterface $theme, $originDir, $targetDir, $symlinkMask)
```
takeit: this method can be removed when https://github.com/Sylius/Sylius/pull/5628 will be merged.

PR #95, opened 2016-08-01
Swp 215

ljubar:
SWP-215 - Expose article meta 'publishedAt'


ljubar:
@ahilles107 please check my PR (SWP-215)

PR #96, opened 2016-08-01
As a user I want to have the ability to switch between many sites within a single organization

takeit:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | yes
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | [SWP-165](https://dev.sourcefabric.org/browse/SWP-165), [SWP-214](https://dev.sourcefabric.org/browse/SWP-214)
| License       | AGPLv3

TODO:
- [x] - [SWP-214](https://dev.sourcefabric.org/browse/SWP-214)
- [x] - [SWP-165](https://dev.sourcefabric.org/browse/SWP-165)
- [x] - update multitenancy bundle docs


src/SWP/Bundle/ContentBundle/Tests/ArticleLoaderTest.php
```diff
@@ -53,11 +49,11 @@ public function testFindNewArticle()
         $this->assertTrue($articleLoader->isSupported('articles'));
         $this->assertFalse($articleLoader->isSupported('items'));
 
-        $article = $articleLoader->load('article', ['contentPath' => '/swp/default/content/test-article']);
+        $article = $articleLoader->load('article', ['contentPath' => '/swp/123456/123abc/content/test-article']);
```
djbrd-sourcefabric: Can you put the path in a variable and reuse it, instead of copy and pasting the value? Then you would only ever have to change it once.

src/SWP/Bundle/ContentBundle/Tests/Controller/RouteControllerTest.php
```diff
@@ -51,12 +48,13 @@ public function testCreateEmptyContentRoutesApi()
         ]);
 
         $this->assertEquals(201, $client->getResponse()->getStatusCode());
-        $this->assertArraySubset(json_decode('{"id":"\/swp\/default\/routes\/simple-test-route","content":null,"static_prefix":null,"variable_pattern":null,"name":"simple-test-route","children":[],"id_prefix":"\/swp\/default\/routes","template_name":null,"type":"content","_links":{"self":{"href":"\/api\/v1\/content\/routes\/\/simple-test-route"}}}', true), json_decode($client->getResponse()->getContent(), true));
+        $this->assertArraySubset(json_decode('{"id":"\/swp\/123456\/123abc\/routes\/simple-test-route","content":null,"static_prefix":null,"variable_pattern":null,"name":"simple-test-route","children":[],"id_prefix":"\/swp\/123456\/123abc\/routes","template_name":null,"type":"content","_links":{"self":{"href":"\/api\/v1\/content\/routes\/\/simple-test-route"}}}', true), json_decode($client->getResponse()->getContent(), true));
```
djbrd-sourcefabric: Wouldn't it be more readable to test the decoded content? For example, $this->assertEquals($content['content'], null), etc. Maintaining these long strings is a pain.

src/SWP/Bundle/ContentBundle/Tests/Controller/RouteControllerTest.php
```diff
@@ -73,7 +71,7 @@ public function testCreateContentRoutesApi()
 
         $this->assertEquals(201, $client->getResponse()->getStatusCode());
         $content = json_decode($client->getResponse()->getContent(), true);
-        $this->assertArraySubset(json_decode('{"id":"\/swp\/default\/routes\/simple-test-route","content":{"id":"\/swp\/default\/content\/test-content-article","title":"Test content article"},"static_prefix":null,"variable_pattern":null,"name":"simple-test-route","children":[],"id_prefix":"\/swp\/default\/routes","template_name":null,"type":"content","_links":{"self":{"href":"\/api\/v1\/content\/routes\/\/simple-test-route"}}, "cache_time_in_seconds":1}', true), $content);
+        $this->assertArraySubset(json_decode('{"id":"\/swp\/123456\/123abc\/routes\/simple-test-route","content":{"id":"\/swp\/123456\/123abc\/content\/test-content-article","title":"Test content article"},"static_prefix":null,"variable_pattern":null,"name":"simple-test-route","children":[],"id_prefix":"\/swp\/123456\/123abc\/routes","template_name":null,"type":"content","_links":{"self":{"href":"\/api\/v1\/content\/routes\/\/simple-test-route"}}, "cache_time_in_seconds":1}', true), $content);
```
djbrd-sourcefabric: Again, this same hard coded value is copied and pasted n times in the same file. In a variable, or a const value, it would be far easier to maintain.

src/SWP/Bundle/FixturesBundle/Command/ThemeSetupCommand.php
```diff
@@ -147,21 +157,24 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
         if (null === $defaultTenant->getThemeName()) {
             $defaultTenant->setThemeName(self::DEFAULT_THEME_NAME);
-            $em = $this->getContainer()->get('doctrine.orm.entity_manager');
-            $em->flush();
+            $documentManager = $this->getContainer()->get('doctrine_phpcr.odm.document_manager');
+            $documentManager->flush();
+        }
+    }
+
+    private function assertTenantIsFound(ThemeAwareTenantInterface $defaultTenant)
```
djbrd-sourcefabric: No need to put this in a function if it's only called once.

src/SWP/Bundle/ContentBundle/Tests/Controller/RouteControllerTest.php
takeit: it's expected value in the test case, not sure if it's so important to make part of it as a const and insert it into the json string. Every expected json string is a bit different, changing it to const is not relavant in this case imho. Whenever you want to change tenant code or an organization code again it can be easily replaced by your editor.

src/SWP/Bundle/FixturesBundle/DataFixtures/ORM/LoadWidgetsData.php
```diff
@@ -26,11 +28,14 @@ class LoadWidgetsData extends AbstractFixture implements FixtureInterface, Order
      */
     public function load(ObjectManager $manager)
     {
+        /** @var TenantInterface $tenant */
```
djbrd-sourcefabric: I'm surprised this works. Suggested way of sharing objects between fixtures is to add a reference. See section with tittle "Sharing Objects between fixtures" here: http://symfony.com/doc/current/bundles/DoctrineFixturesBundle/index.html

src/SWP/Bundle/FixturesBundle/DataFixtures/PHPCR/LoadHomepagesData.php
```diff
@@ -0,0 +1,80 @@
...
+use SWP\Component\MultiTenancy\Model\TenantInterface;
+
+class LoadHomepagesData extends AbstractFixture implements FixtureInterface, OrderedFixtureInterface
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function load(ObjectManager $manager)
+    {
+        $tenants = [
```
djbrd-sourcefabric: Again, I think you should use references here.

src/SWP/Bundle/FixturesBundle/DataFixtures/PHPCR/LoadHomepagesData.php
```diff
@@ -0,0 +1,80 @@
...
+            $manager->find(Tenant::class, '/swp/123456/123abc'),
+            $manager->find(Tenant::class, '/swp/654321/456def'),
+        ];
+
+        $this->createHomepages($manager, $tenants);
+
+        foreach ($tenants as $site) {
+            if ($site instanceof TenantInterface) {
```
djbrd-sourcefabric: I guess that you're actually checking for null here? I would do the check where you retrieve the objects, if they shouldn't be null

src/SWP/Bundle/FixturesBundle/Command/ThemeSetupCommand.php
takeit: it's more readable, it moves throwing an exception to another function which is one operation and each method should do only one operation - the execute method could be improved here tho to make it max 20-25 lines.

src/SWP/Bundle/FixturesBundle/DataFixtures/PHPCR/LoadRoutesData.php
```diff
@@ -34,7 +35,7 @@ public function load(ObjectManager $manager)
     {
         $env = $this->getEnvironment();
         if ('test' === $env) {
-            $parent = $manager->find(null, '/swp/default/routes');
+            $parent = $manager->find(null, $this->getTenantPrefix().'/routes');
```
djbrd-sourcefabric: Use a reference

src/SWP/Bundle/FixturesBundle/DataFixtures/ORM/LoadWidgetsData.php
takeit: the fixture is container aware so it allows you to use other services like the repository. Note that the widgets are stored using doctrine orm and tenants are stored in phpcr - the reference here will not work.

src/SWP/Bundle/FixturesBundle/DataFixtures/PHPCR/LoadHomepagesData.php
takeit: References don't work when using data fixtures with PHPCR - see PR I created: https://github.com/doctrine/data-fixtures/pull/241

src/SWP/Bundle/MultiTenancyBundle/Command/CreateOrganizationCommand.php
```diff
@@ -0,0 +1,149 @@
...
+        $organization->setName($name);
+        $organization->setEnabled(!$disabled);
+
+        return $organization;
+    }
+
+    /**
+     * @return ObjectManager
+     */
+    protected function getObjectManager()
```
djbrd-sourcefabric: I'm not so keen on these one line functions, especially if they are only used in one other function.

src/SWP/Bundle/MultiTenancyBundle/DependencyInjection/Compiler/RegisterOrganizationFactoryPass.php
```diff
@@ -0,0 +1,61 @@
...
+use Symfony\Component\DependencyInjection\Definition;
+use Symfony\Component\DependencyInjection\Parameter;
+use Symfony\Component\DependencyInjection\Reference;
+
+class RegisterOrganizationFactoryPass implements CompilerPassInterface
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function process(ContainerBuilder $container)
```
djbrd-sourcefabric: This seems unnecessarily complicated to me for what you're trying achieve here. There is very little logic in this class, and I wonder if the same results could be achieved in yml. Should it actually be possible to create an organization factory with different parameters? Do we need that level of flexibility?

takeit: Yes, it can be achieved with yaml config but I created compiler pass because I can unit test it, with yaml config you can't.

src/SWP/Bundle/MultiTenancyBundle/Doctrine/ORM/OrganizationRepository.php
```diff
@@ -0,0 +1,53 @@
...
+
+use SWP\Bundle\StorageBundle\Doctrine\ORM\EntityRepository;
+use SWP\Component\MultiTenancy\Repository\OrganizationRepositoryInterface;
+
+class OrganizationRepository extends EntityRepository implements OrganizationRepositoryInterface
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function findByName($name)
```
djbrd-sourcefabric: I would rename these methods to findEnabledByName, etc. Unless you always want to get only enabled entities, in which case, you could use a filter.

src/SWP/Bundle/MultiTenancyBundle/Doctrine/PHPCR/OrganizationRepository.php
```diff
@@ -0,0 +1,50 @@
...
+namespace SWP\Bundle\MultiTenancyBundle\Doctrine\PHPCR;
+
+use SWP\Bundle\StorageBundle\Doctrine\ODM\PHPCR\DocumentRepository;
+use SWP\Component\MultiTenancy\Repository\OrganizationRepositoryInterface;
+
+class OrganizationRepository extends DocumentRepository implements OrganizationRepositoryInterface
+{
+    /**
+     * {@inheritdoc}
+     */
```
djbrd-sourcefabric: Might be easier to use SoftDeleteable functionality, for which there is already a filter. Deleted entities can be restored. So there is effectively no difference between disabling an entity and deleting it.

takeit: SoftDeleteable is not supported in PHPCR.

src/SWP/Bundle/MultiTenancyBundle/EventListener/TenantSubscriber.php
```diff
@@ -75,8 +76,10 @@ protected function addTenant(LifecycleEventArgs $args)
                 return;
             }
 
-            if (null === $this->tenantContext) {
-                $this->tenantContext = $this->container->get('swp_multi_tenancy.tenant_context');
+            if ($this->phpcrEnabled) {
```
djbrd-sourcefabric: I wonder why it's necessary to set the code of the tenant if phpcr is enabled. Looks weird to call a setter with different types of argument.

src/SWP/Bundle/MultiTenancyBundle/EventListener/TenantableListener.php
```diff
@@ -53,12 +53,12 @@ public function __construct(EntityManagerInterface $entityManager, TenantContext
...
     {
         $tenant = $this->tenantContext->getTenant();
+
         if ($tenant && $tenant->getId()) {
-            $tenantId = $tenant->getId();
             $this->entityManager
                 ->getFilters()
                 ->enable('tenantable')
-                ->setParameter('tenantId', $tenantId);
+                ->setParameter('tenantId', $tenant->getCode());
```
djbrd-sourcefabric: Seems weird to set a parameter called 'tenantId' with the result of getCode, if the same object has a getId method

src/SWP/Bundle/MultiTenancyBundle/EventListener/TenantSubscriber.php
takeit:
```When PHPCR ODM persistence backend is enabled it will rely on tenant's unique code instead of the tenant id.
In this case, if the tenant exists in the context and the tenant code is 123abc, it will add ``WHERE tenant_id = 123abc`` to every select query.```

If you are using ORM for example, you would want to make a relation to tenant. If you use phpcr and you want to make some entities tenant aware you cant make a relations between ORM and PHPCR that's why a code is used.

File: src/SWP/Bundle/MultiTenancyBundle/Factory/OrganizationFactory.php [hunk 15]
```diff
@@ -0,0 +1,75 @@
...
+    ) {
+        $this->decoratedFactory = $decoratedFactory;
+        $this->documentManager = $documentManager;
+        $this->rootPath = $rootPath;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function create()
```
djbrd-sourcefabric: So it should be possible to create an organization without a code? If not, I think that this method shouldn't be here.

src/SWP/Bundle/MultiTenancyBundle/Resources/config/doctrine-orm/Tenant.orm.yml
```diff
@@ -0,0 +1,52 @@
...
+        deletedAt:
+            type: datetime
+            nullable: true
+    manyToOne:
+        organization:
+            targetEntity: SWP\Component\MultiTenancy\Model\Organization
+            inversedBy: tenants
+            cascade: ["remove"]
+            joinColumn:
+                name: organization_id
```
djbrd-sourcefabric: No point explicitly including defaults, I think

src/SWP/Component/Common/Generator/RandomStringGenerator.php
```diff
@@ -0,0 +1,58 @@
...
+ *
+ * For the full copyright and license information, please see the
+ * AUTHORS and LICENSE files distributed with this source code.
+ *
+ * @copyright 2015 Sourcefabric z.ú.
+ * @license http://www.superdesk.org/license
+ */
+namespace SWP\Component\Common\Generator;
+
+class RandomStringGenerator implements GeneratorInterface
```
djbrd-sourcefabric: Any reason for writing bespoke code for this, and not using a library function, like uniqid?

src/SWP/Bundle/MultiTenancyBundle/SWPMultiTenancyBundle.php
```diff
@@ -27,5 +29,25 @@ public function build(ContainerBuilder $container)
         $container->addCompilerPass(new ConfigurePrefixCandidatesCompilerPass());
         $container->addCompilerPass(new TenantAwareRouterCompilerPass());
         $container->addCompilerPass(new RegisterTenantFactoryCompilerPass());
+        $container->addCompilerPass(new RegisterOrganizationFactoryPass());
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getSupportedDrivers()
```
djbrd-sourcefabric: Are you sure the bundle works with the Doctrine ORM driver? If we're going to support this, tests should probably be written

takeit: no it doesnt atm, with next iteration I will look at it - see attached docs.

src/SWP/Bundle/MultiTenancyBundle/Tests/Command/CreateOrganizationCommandTest.php
```diff
@@ -0,0 +1,163 @@
...
+     * @covers SWP\Bundle\MultiTenancyBundle\Command\CreateOrganizationCommand
+     */
+    public function testExecuteDisabledOrganization()
+    {
+        $this->dialog->setInputStream($this->getInputStream("Example\n"));
+        $this->command->setContainer($this->getMockContainer(null, new Organization(), 'Example'));
+        $this->commandTester = new CommandTester($this->command);
+        $this->commandTester->execute([
+            'command' => $this->command->getName(),
+            '--disabled' => true,
```
djbrd-sourcefabric: Seems strange to be able to disable an organization with a command called swp:organization:create and a flag. I would write a swp:organization:disable command. The two commands could derive from a common base class in order to share functionality.

src/SWP/Component/Common/Generator/RandomStringGenerator.php
takeit: we can generate whatever code length we want, we dont need over 10 chars codes which uniqid generates. This way we can customize it by having a custom generator.

src/SWP/Bundle/MultiTenancyBundle/Tests/Command/CreateTenantCommandTest.php
```diff
@@ -93,29 +113,36 @@ public function testExecuteWhenDefaultTenantExists()
...
-    private function getMockContainer($mockTenant = null, $subdomain = 'default')
+    private function getMockContainer($mockTenant = null, $mockOrganization = null, $mockedTenantInFactory = null, $subdomain = 'default')
     {
-        $mockRepo = $this->getMockBuilder('SWP\Component\MultiTenancy\Repository\TenantRepositoryInterface')
+        $mockRepoOrganization = $this->getMockBuilder(OrganizationRepositoryInterface::class)
+            ->getMock();
+
+        $mockRepoOrganization->expects($this->any())
+            ->method('findByCode')
+            ->with('123456')
```
djbrd-sourcefabric: Value should be in a const, I think

src/SWP/Bundle/MultiTenancyBundle/spec/DependencyInjection/Compiler/RegisterOrganizationFactoryPassSpec.php
```diff
@@ -0,0 +1,111 @@
+<?php
```
djbrd-sourcefabric: I honestly cannot see the point of these spec tests. The only time any one of them has failed is when I've purposefully changed something. I can maybe see the point in writing them first, so you define the behaviour before actually coding it, but if there is a reason to write them afterwards, I'd love to know what it is.

src/SWP/Bundle/TemplateEngineBundle/Model/Container.php
```diff
@@ -77,7 +76,7 @@ class Container implements ContainerInterface, TenantAwareInterface, Timestampab
     protected $widgets;
 
     /**
-     * @var TenantInterface
+     * @var string
```
djbrd-sourcefabric: If this is a tenant code, and not a tenant, please change the name of the variable. I don't know why you would to do this, and not create relation, or leave the relation actually.

src/SWP/Bundle/TemplateEngineBundle/Model/WidgetModel.php
```diff
@@ -65,7 +64,7 @@ class WidgetModel implements WidgetModelInterface, TenantAwareInterface, Timesta
     protected $containers;
 
     /**
-     * @var TenantInterface
+     * @var string
```
djbrd-sourcefabric: Again, why a string here and not a relation? So that you can mix ORM and ODM?

src/SWP/Bundle/TemplateEngineBundle/Resources/config/doctrine/Container.orm.yml
```diff
@@ -42,14 +42,11 @@ SWP\Bundle\TemplateEngineBundle\Model\Container:
             gedmo:
                 timestampable:
                     on: update
+        tenant:
```
djbrd-sourcefabric: Why do you want to bypass doctrine? I really cannot understand this.

src/SWP/Bundle/WebRendererBundle/Controller/DefaultController.php
```diff
@@ -25,12 +26,12 @@ class DefaultController extends Controller
...
-        $pathBuilder = $this->get('swp_multi_tenancy.path_builder');
-        $site = $this->get('swp.repository.site')->find($pathBuilder->build('/'));
-        $homepage = $site->getHomepage();
+        /** @var HomepageBasedTenantInterface $currentTenant */
+        $currentTenant = $this->get('swp_multi_tenancy.tenant_context')->getTenant();
+        $homepage = $currentTenant->getHomepage();
 
-        if (null === $homepage) {
-            throw $this->createNotFoundException('No homepage configured!');
+        if (null !== $homepage) {
```
djbrd-sourcefabric: So it should be possible for a tenant not to have a homepage? What is the behaviour?

src/SWP/Component/Common/Generator/GeneratorInterface.php
```diff
@@ -0,0 +1,22 @@
+<?php
+
```
djbrd-sourcefabric: 'Generator' is a bit of a generic name, I think. What does it generate? Anything? I can't imagine how such a generic interface could be useful.

src/SWP/Bundle/WebRendererBundle/Controller/DefaultController.php
takeit: yes, it fallbacks to default index.html.twig in theme currently

src/SWP/Component/MultiTenancy/Factory/TenantFactory.php
```diff
@@ -13,31 +13,73 @@
...
-     * @param string $className
+     * @param FactoryInterface                $decoratedFactory
+     * @param GeneratorInterface              $generator
+     * @param OrganizationRepositoryInterface $organizationRepository
      */
-    public function __construct($className)
-    {
-        $this->className = $className;
+    public function __construct(
+        FactoryInterface $decoratedFactory,
```
djbrd-sourcefabric: Shouldn't this be a TenantFactoryInterface?

src/SWP/Component/MultiTenancy/Factory/TenantFactory.php
```diff
@@ -13,31 +13,73 @@
...
+        $this->organizationRepository = $organizationRepository;
     }
 
     /**
      * {@inheritdoc}
      */
     public function create()
     {
-        return new $this->className();
+        /** @var TenantInterface $tenant */
```
djbrd-sourcefabric: Do you want to enable the creation of a Tenant without an Organization?

src/SWP/Component/MultiTenancy/Model/AbstractOrganization.php
```diff
@@ -0,0 +1,90 @@
+<?php
```
djbrd-sourcefabric: And it should also be possible to have an organization without a tenant? This code is flexible, but unless all permutations are actually supported, you're really just making the code unnecessarily complex, I think.

src/SWP/Component/MultiTenancy/Factory/TenantFactory.php
takeit: This is component so - yes, it should be possible as it can be used not only in SWP

src/SWP/Component/MultiTenancy/Model/OrganizationInterface.php
```diff
@@ -0,0 +1,84 @@
...
+ * @license http://www.superdesk.org/license
+ */
+namespace SWP\Component\MultiTenancy\Model;
+
+use SWP\Component\Common\Model\EnableableInterface;
+use SWP\Component\Common\Model\SoftDeletableInterface;
+use SWP\Component\Common\Model\TimestampableInterface;
+use SWP\Component\Storage\Model\PersistableInterface;
+
+interface OrganizationInterface extends PersistableInterface, SoftDeletableInterface, EnableableInterface, TimestampableInterface
```
djbrd-sourcefabric: By this definition, an organization has to have a tenant.

src/SWP/Component/MultiTenancy/Model/TenantInterface.php
```diff
@@ -66,4 +66,24 @@ public function getName();
      * @param string $name The tenant name
```
djbrd-sourcefabric: And, by this definition, a Tenant has to be related to an organization.

src/SWP/Component/MultiTenancy/Model/AbstractOrganization.php
takeit: what do you suggest then?

src/SWP/Component/MultiTenancy/spec/PathBuilder/TenantAwarePathBuilderSpec.php
```diff
@@ -41,16 +47,21 @@ public function it_implements_path_builder_interface()
...
+        $this->build('/routes/articles')->shouldReturn('/swp/123456/654321');
+        $this->build('routes')->shouldReturn('/swp/123456/654321/routes');
+        $this->build('/')->shouldReturn('/swp/123456/654321');
         $this->build('routes', 'context')->shouldReturn('/swp/context/routes');
     }
 
     public function it_should_throw_an_exception_when_no_tenant_and_empty_path($tenantContext)
     {
-        $tenantContext->getTenant()->willReturn(new Tenant());
+        $currentTenant = new Tenant();
```
djbrd-sourcefabric: Maybe put these 6 lines in a function, as they're copied twice below

src/SWP/Component/Common/Generator/GeneratorInterface.php
takeit: You can have also other generators like random number generator etc. then I see no problem with GeneratorInterface name.

src/SWP/Component/MultiTenancy/Factory/TenantFactory.php
takeit: no, its Factory from storage component which is being decorated can create you a pure object based on the provided class name in the construct. See storage bundle and storage component documentation. TenantFactory decorates the default one so you can create a tenant with code or with organization.

src/SWP/Component/MultiTenancy/Model/AbstractOrganization.php
djbrd-sourcefabric: There is no need for an AbstractOrganization class, I think. Just an Organization class. If it's not possible to have an Organization without a tenant.

src/SWP/Bundle/ContentBundle/Tests/Controller/RouteControllerTest.php
djbrd-sourcefabric: I think that these long lines of json are a pain to read and maintain.

src/SWP/Bundle/FixturesBundle/Command/ThemeSetupCommand.php
djbrd-sourcefabric: Yeah, the execute method does two very different things - deleting and setting up a theme. If we're going to put any code that might throw an exception into its own method, we will have to do a lot of refactoring.

src/SWP/Bundle/FixturesBundle/DataFixtures/ORM/LoadWidgetsData.php
djbrd-sourcefabric: Tenants were stored using doctrine orm before. Any reason for the change? I think that mixing the two is not such a great idea actually. It is obviously not explicitly supported.

src/SWP/Bundle/FixturesBundle/DataFixtures/PHPCR/LoadHomepagesData.php
djbrd-sourcefabric: Quite a lot of things don't work with PHPCR, it seems. Makes me wonder whether it's a good idea to use it.

src/SWP/Bundle/FixturesBundle/DataFixtures/ORM/LoadWidgetsData.php
takeit: That's why we rely on tenant's code instead of relationship. It was changed to PHPCR because we were duplicating the tenants in phpcr. Tenants were stored using doctrine orm and when we initialized phpcr tenants were also inserted (documents) based on the tenants which makes no sense. We kept the tenants in orm and just a tenant document with id in phpcr. Now we keep all in PHPCR and rely on tenant's code when using entities.

src/SWP/Bundle/TemplateEngineBundle/Model/Container.php
takeit: This TemplateEngineBundle supports only Doctrine ORM entiries. To solve it properly, a Container class should be created in `WebRendererBundle` (I will rename it in next PR to `CoreBundle`) which extends `TemplateEngineBundle/Model/Container.php` and in this new class we should add the `TenantAwareInterface`, override the mapping file for Container class to use tenant code. This way we will decouple MultiTenancy from `TemplateEngineBundle`.

src/SWP/Bundle/ContentBundle/Tests/Controller/RouteControllerTest.php
takeit: its easier to catch issues if you assert exact json string, array subset was used here to skip fields with datetime.

src/SWP/Bundle/MultiTenancyBundle/Tests/Command/CreateOrganizationCommandTest.php
takeit: It just the flag indicating which status should have a newly created organization, either it should be enabled or disabled by default. You can't disable/enable existing organizations so I think that a separate command is not needed. If you would want to disable/enable existing organizations then it makes sense to create `swp:organization:disable` and `swp:organization:enable` commands.

src/SWP/Bundle/MultiTenancyBundle/spec/DependencyInjection/Compiler/RegisterOrganizationFactoryPassSpec.php
takeit: that's the whole idea of the phpspec tool - you should first write specs describing the behaviour before the implementation.

src/SWP/Component/MultiTenancy/Model/TenantInterface.php
takeit: it is, see `$organization` property

src/SWP/Component/MultiTenancy/Model/OrganizationInterface.php
takeit: no, it doesn't have to. You can have an empty array. It doesnt tell you you must have at least one tenant.

src/SWP/Bundle/MultiTenancyBundle/Command/CreateOrganizationCommand.php
ahilles107: Thanks to this you can change used object_manager with your own implementation with replacing just one method. It's good practice.

src/SWP/Bundle/MultiTenancyBundle/Doctrine/ORM/OrganizationRepository.php
ahilles107: filter sounds like good idea

src/SWP/Bundle/MultiTenancyBundle/EventListener/TenantableListener.php
ahilles107: How about changing `tenantId` in orm to `tenantCode`?

src/SWP/Bundle/MultiTenancyBundle/SWPMultiTenancyBundle.php
ahilles107: So we should remove for now `Drivers::DRIVER_DOCTRINE_ORM`

ahilles107:
👍

PR #100, opened 2016-08-02
Revisions - enables user to create clone of a container and of a widget, and to copy a those clones back to their sources after they have been changed

djbrd-sourcefabric:
| Q | A |
| --- | --- |
| Bug fix? | [no] |
| New feature? | [yes] |
| BC breaks? | [no] |
| Deprecations? | [no] |
| Tests pass? | [yes] |
| Fixed tickets | [SWP-207, SWP-208] |
| License | AGPLv3 |

Added endpoints to the api to enable cloning of widgets and containers, and then copying them back to their originals from which they were cloned, after changing them.


src/SWP/Bundle/TemplateEngineBundle/Form/Type/ContainerSelectorType.php
```diff
@@ -0,0 +1,12 @@
+<?php
+/**
```
ahilles107: wrong doc block

src/SWP/Bundle/TemplateEngineBundle/Service/RevisionService.php
```diff
@@ -0,0 +1,132 @@
...
+class RevisionService
+{
+    const REVISION_NAME_SEPARATOR = '_';
+
+    /**
+     * @var ObjectManager
+     */
+    protected $objectManager;
+
+    public function __construct(ObjectManager $objectManager)
```
ahilles107: docblocks are missing

src/SWP/Bundle/TemplateEngineBundle/Service/RevisionService.php
```diff
@@ -0,0 +1,132 @@
...
+        $published->setName($published->getName().self::REVISION_NAME_SEPARATOR.uniqid());
+
+        $unpublished->onPublished($published);
+
+        $this->objectManager->flush();
+
+        return $unpublished;
+    }
+
+    public function getCurrentRevisions($className)
```
ahilles107: It can return You many revisions here? I think that you should use `findOneBy` and change name to `getCurrentRevision`.

PR #102, opened 2016-08-08
Fix of gimmelist block in twig with limit, offset and order by parameters

djbrd-sourcefabric:
| Q             | A
| ------------- | ---
| Bug fix?      | [yes]
| New feature?  | [no]
| BC breaks?    | [no]
| Deprecations? | [no]
| Tests pass?   | [yes]
| Fixed tickets | [SWP-236]
| License       | AGPLv3

### DIFF
--- a/src/SWP/Bundle/ContentBundle/Loader/ArticleLoader.php
+++ b/src/SWP/Bundle/ContentBundle/Loader/ArticleLoader.php
@@ -137,7 +137,7 @@ class ArticleLoader implements LoaderInterface
                             throw new \Exception('Order filter must have two parameters with second one asc or desc, e.g. order(id, desc)');
                         }
                         if ($order[0] === 'id') {
-                            $order[0] = 'uuid';
+                            $order[0] = 'jcr:uuid';
                         }
                         $queryStr .= ' ORDER BY S.'.$order[0].' '.$order[1];
                     }
--- a/src/SWP/Bundle/ContentBundle/Loader/ArticleLoader.php
+++ b/src/SWP/Bundle/ContentBundle/Loader/ArticleLoader.php
@@ -136,7 +136,7 @@ class ArticleLoader implements LoaderInterface
                     if (isset($parameters['order'])) {
                         $order = $parameters['order'];
                         if (!is_array($order) || count($order) !== 2 || (strtoupper($order[1]) != 'ASC' && strtoupper($order[1]) != 'DESC')) {
-                            throw new \Exception('Order filter must have two parameters with second once asc or desc');
+                            throw new \Exception('Order filter must have two parameters with second one asc or desc, e.g. order(id, desc)');
                         }
                         if ($order[0] === 'id') {
                             $order[0] = 'uuid';
@@ -149,9 +149,11 @@ class ArticleLoader implements LoaderInterface
                     if (isset($parameters['limit'])) {
                         $query->setLimit($parameters['limit']);
                     }
+
                     if (isset($parameters['start'])) {
                         $query->setOffset($parameters['start']);
                     }
+
                     $articles = $this->dm->getDocumentsByPhpcrQuery($query);
 
                     $metas = [];
--- a/src/SWP/Bundle/ContentBundle/Loader/ArticleLoader.php
+++ b/src/SWP/Bundle/ContentBundle/Loader/ArticleLoader.php
@@ -13,9 +13,7 @@
  */
 namespace SWP\Bundle\ContentBundle\Loader;
 
-use Doctrine\Common\Collections\Criteria;
 use PHPCR\Query\QueryInterface;
-use PHPCR\Util\QOM\QueryBuilder;
 use SWP\Component\TemplatesSystem\Gimme\Loader\LoaderInterface;
 use SWP\Component\TemplatesSystem\Gimme\Meta\Meta;
 use Symfony\Component\Yaml\Parser;
--- a/src/SWP/Bundle/ContentBundle/Loader/ArticleLoader.php
+++ b/src/SWP/Bundle/ContentBundle/Loader/ArticleLoader.php
@@ -131,7 +131,7 @@ class ArticleLoader implements LoaderInterface
                     $node = $this->dm->getNodeForDocument($route);
                     $identifier = $node->getIdentifier();
 
-                    $queryStr = "SELECT * FROM [nt:unstructured] as S ";
+                    $queryStr = 'SELECT * FROM [nt:unstructured] as S ';
                     $queryStr .= "WHERE S.phpcr:class='SWP\Bundle\ContentBundle\Doctrine\ODM\PHPCR\Article' ";
                     $queryStr .= "AND S.route=$identifier";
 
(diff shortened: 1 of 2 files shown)


src/SWP/Bundle/ContentBundle/Loader/ArticleLoader.php
```diff
@@ -125,7 +126,34 @@ public function load($type, $parameters, $responseType = LoaderInterface::SINGLE
...
+                    $identifier = $node->getIdentifier();
+
+                    $queryStr = 'SELECT * FROM [nt:unstructured] as S ';
+                    $queryStr .= "WHERE S.phpcr:class='SWP\Bundle\ContentBundle\Doctrine\ODM\PHPCR\Article' ";
+                    $queryStr .= "AND S.route=$identifier";
+
+                    if (isset($parameters['order'])) {
+                        $order = $parameters['order'];
+                        if (!is_array($order) || count($order) !== 2 || (strtoupper($order[1]) != 'ASC' && strtoupper($order[1]) != 'DESC')) {
+                            throw new \Exception('Order filter must have two parameters with second once asc or desc');
```
takeit: I don't understand this exception  msg - `..with second once..`

src/SWP/Bundle/ContentBundle/Loader/ArticleLoader.php
```diff
@@ -125,7 +126,34 @@ public function load($type, $parameters, $responseType = LoaderInterface::SINGLE
                 ));
 
                 if ($route) {
-                    $articles = $this->dm->getReferrers($route, null, null, null, 'SWP\Bundle\ContentBundle\Doctrine\ODM\PHPCR\Article');
+                    $node = $this->dm->getNodeForDocument($route);
+                    $identifier = $node->getIdentifier();
+
+                    $queryStr = 'SELECT * FROM [nt:unstructured] as S ';
+                    $queryStr .= "WHERE S.phpcr:class='SWP\Bundle\ContentBundle\Doctrine\ODM\PHPCR\Article' ";
+                    $queryStr .= "AND S.route=$identifier";
```
takeit: I would use `sprintf` function to construct that query

src/SWP/Bundle/ContentBundle/Loader/ArticleLoader.php
```diff
@@ -125,7 +126,34 @@ public function load($type, $parameters, $responseType = LoaderInterface::SINGLE
...
+                    }
+
+                    $query = $this->dm->createPhpcrQuery($queryStr, QueryInterface::JCR_SQL2);
+
+                    if (isset($parameters['limit'])) {
+                        $query->setLimit($parameters['limit']);
+                    }
+                    if (isset($parameters['start'])) {
+                        $query->setOffset($parameters['start']);
+                    }
```
takeit: new line

src/SWP/Bundle/ContentBundle/Loader/ArticleLoader.php
```diff
@@ -125,7 +126,34 @@ public function load($type, $parameters, $responseType = LoaderInterface::SINGLE
...
+                            $order[0] = 'uuid';
+                        }
+                        $queryStr .= ' ORDER BY S.'.$order[0].' '.$order[1];
+                    }
+
+                    $query = $this->dm->createPhpcrQuery($queryStr, QueryInterface::JCR_SQL2);
+
+                    if (isset($parameters['limit'])) {
+                        $query->setLimit($parameters['limit']);
+                    }
```
takeit: new line

ahilles107:
Please add new tests for this feature. We need to know when it breaks.

src/SWP/Bundle/ContentBundle/Loader/ArticleLoader.php
djbrd-sourcefabric: Easier to do it without, I think. sprintf is from C.

src/SWP/Bundle/ContentBundle/Loader/ArticleLoader.php
djbrd-sourcefabric: Style is already very rigid. I can't really understand the point of insisting on new lines when even Styleci doesn't.

ahilles107: We try to keep code standards close to Symfony (http://symfony.com/doc/current/contributing/code/standards.html), and there is always new line after if block (as long as it's not the last one thing in function). In rest of file there is new line after if block - why here we should skip it?

src/SWP/Bundle/ContentBundle/Loader/ArticleLoader.php
ahilles107: sprintf is in php (http://php.net/manual/en/function.sprintf.php) also. And it's preferred for passing parameters to strings.

ahilles107: And it think that whole discussion here is not needed as You should use `setParameter/setParameters` on `Query` object. 
```
$query = $this->dm->createPhpcrQuery($queryStr, QueryInterface::JCR_SQL2);
$query->setParameter('identifier', $identifier);
```

ahilles107: @djbrd-sourcefabric - can you fix it?

djbrd-sourcefabric: The query class has no setParameter method. The setParameter method is in the queryBuilder class. I could the query builder to build the query, but then I'd have to work out how to generate the query which is already there using the query builder, which seems like a waste of time. Seeing as the value of the parameter is provided by us, and not by an end user, I think we can safely put it directly in the query without using a placeholder. If you disagree, let me know, and I'll work out how to use the PhpcrQueryBuilder, but this will probably require trial and error because I can't find documentation on this query builder, only on the one returned by $this->dm->createQueryBuilder() which doesn't enable this kind of a query.

ahilles107: Result of this method is not `Doctrine\ODM\PHPCR\Query\Query`? (https://github.com/doctrine/phpcr-odm/blob/241d3622e1a348e5fabfadc7a4195673c472338c/lib/Doctrine/ODM/PHPCR/Query/Query.php#L110)

ahilles107: It should: https://github.com/doctrine/phpcr-odm/blob/master/lib/Doctrine/ODM/PHPCR/DocumentManagerInterface.php#L197

djbrd-sourcefabric: No it's Jackalope\Query\SqlQuery. The reason that the code is so funky is that with Doctrine's implementation we cannot create such a query (finding all documents with a reference to another document, and with some other criteria).

ahilles107: Ok, so we wait for tests.

djbrd-sourcefabric:
Tests are added

PR #103, opened 2016-08-09
Package and related article overwritten when a package with the same guid, or an article with the same slug is pushed

djbrd-sourcefabric:
| Q             | A
| ------------- | ---
| Bug fix?      | [yes]
| New feature?  | [no]
| BC breaks?    | [no]
| Deprecations? | [no]
| Tests pass?   | [yes]
| Fixed tickets | [SWP-226, SWP-240]
| License       | AGPLv3

When a package is pushed, the db is checked for an existing package with the same guid. If there is one, this package is deleted (but can be retrieved, because a package is soft deletable). If there already exists an article with the same slug as the one in the package, this article is deleted. Articles are currently not soft deletable.
Items are also now stored in a separate table, as an exception was thrown when trying to retrieve them in serialized form from the db.
Use of Pipeline in pushContentAction looked superfluous.


src/SWP/Bundle/ContentBundle/Controller/ContentPushController.php
```diff
@@ -43,23 +43,28 @@ class ContentPushController extends FOSRestController
...
-            ->pipe([$this->get('swp_content.transformer.package_to_article'), 'transform'])
-            ->pipe(function ($article) {
-                $this->get('swp.repository.article')->add($article);
-                $this->get('event_dispatcher')->dispatch(ArticleEvents::POST_CREATE, new ArticleEvent($article));
-
-                return $article;
-            });
-
-        $pipeline->process($request->getContent());
+        $content = $request->getContent();
```
ahilles107: Why you resigned from Pipeline solution?

src/SWP/Bundle/BridgeBundle/Resources/config/doctrine-orm/Item.orm.yml
```diff
@@ -0,0 +1,54 @@
+SWP\Component\Bridge\Model\Item:
```
takeit: why do you need to keep item in a separate table? can't it be in one table (swp_package) as a serialized array collection? Serializer would then automatically handle it, and you won't need to fetch data based on the relation.

src/SWP/Bundle/ContentBundle/Controller/ContentPushController.php
djbrd-sourcefabric: Couldn't understand need for the pipeline. What does it add? And it makes branching more difficult. I could put it back in, but if there is no reason for it, then it just makes the code less readable.

src/SWP/Bundle/BridgeBundle/Resources/config/doctrine-orm/Item.orm.yml
djbrd-sourcefabric: There is no out of the box solution for this. We would have to create a custom mapping type. The existing Doctrine mapping types for array are ArrayType, which calls serialize and unserialize on the array, SimpleArrayType, which calls implode and explode, and JsonArrayType, which calls json_encode and json_decode. The Doctrine way of storing arrays of objects within an object is to store those objects in a separate table with a foreign key reference. The collection is only hydrated when the objects are required. Migration is supported. Serializing and deserializing an array is not an utterly trivial operation. So, I doubt that flattening an array of objects and storing it in one column is even optimal anyway.

src/SWP/Bundle/ContentBundle/Controller/ContentPushController.php
ahilles107: Personally i prefer Pipline solution - it's more readable and organized. We emit here events - just use them for removing already existing packages and articles.

takeit:
remove `league/pipeline` from composer.json then if the pipeline is remove.

src/SWP/Bundle/ContentBundle/Controller/ContentPushController.php
takeit: same here, for me its more organized and the library says its all: `The pipeline pattern allows you to easily compose sequential stages by chaining stages.` thought is better to put it into a logical sequence instead. But we can resign from it for now - that's not a big deal.

djbrd-sourcefabric: I still really can't see any benefit in using the pipeline over just calling the functions sequentially. A programming language enables putting stages into functions and calling them one by one. Without any apparent benefit, the use of pipeline looks like pure over-engineering to, but if there is actually a genuine benefit, let's restore it.

PR #109, opened 2016-08-16
Implemented twig extension for Strings library

djbrd-sourcefabric:
| Q             | A
| ------------- | ---
| Bug fix?      | [yes]
| New feature?  | [yes]
| BC breaks?    | [no]
| Deprecations? | [no]
| Tests pass?   | [yes]
| Fixed tickets | [SWP-235]
| License       | AGPLv3

All functions in this Stringy library (https://github.com/danielstjules/Stringy) have been made available in twig. Functions which manipulate strings are available as filters, while functions which return booleans are available as functions. So, {{ value|camelize }} will render the value in camel case, while {% if hasUpperCase(value) %} will be true if the value has any upper case character.

### DIFF
--- a/src/SWP/Bundle/WebRendererBundle/Twig/StringyExtension.php
+++ b/src/SWP/Bundle/WebRendererBundle/Twig/StringyExtension.php
@@ -1,0 +1,90 @@
+<?php
+
+/**
+ * This file is part of the Superdesk Web Publisher Template Engine Bundle.
+ *
+ * Copyright 2016 Sourcefabric z.ú. and contributors.
+ *
+ * For the full copyright and license information, please see the
+ * AUTHORS and LICENSE files distributed with this source code.
+ *
+ * @copyright 2016 Sourcefabric z.ú.
+ * @license http://www.superdesk.org/license
+ */
+namespace SWP\Bundle\WebRendererBundle\Twig;
+
+class StringyExtension extends \Twig_Extension
+{
+    const EXCLUDE_FUNCTIONS = ['__construct', '__toString', 'create'];
+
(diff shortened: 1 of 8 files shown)


ahilles107:
Looks great! This is what we wanted ;) Please add some docs.

src/SWP/Bundle/WebRendererBundle/Twig/StringyExtension.php
```diff
@@ -0,0 +1,88 @@
...
+            }
+
+            if ($this->environment->getFilter($name)) {
+                continue;
+            }
+
+            $method = $stringyClass->getMethod($name);
+            $doc = $method->getDocComment();
+            if (strpos($doc, '@return bool')) {
+                $this->functions[$name] = new \Twig_SimpleFunction($name, array('Stringy\StaticStringy', $name));
```
takeit: let's use new array syntax `[]` instead old one: `array()` - `[StaticStringy::class, $name]`

src/SWP/Bundle/WebRendererBundle/Twig/StringyExtension.php
```diff
@@ -0,0 +1,88 @@
...
+{
+    const EXCLUDE_FUNCTIONS = ['__construct', '__toString', 'create'];
+
+    protected $environment;
+
+    protected $functions = [];
+
+    protected $filters = [];
+
+    public function __construct(\Twig_Environment $environment)
```
takeit: doc blocks are missing here and for properties

takeit:
waiting for docs on this feature, besides that 👍

PR #115, opened 2016-08-23
Allow to read/write article metadata based on provided package

takeit:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | [SWP-188](https://dev.sourcefabric.org/browse/SWP-188)
| License       | AGPLv3


src/SWP/Bundle/ContentBundle/Resources/config/doctrine-phpcr/Article.phpcr.yml
```diff
@@ -35,6 +35,9 @@ SWP\Bundle\ContentBundle\Doctrine\ODM\PHPCR\Article:
             nullable: true
         isPublishable:
             type: boolean
+        metadata:
```
djbrd-sourcefabric: Is there not an issue here with array to string conversion? Looks like metadata is an array in the class, and configured here as a string

src/SWP/Bundle/ContentBundle/EventListener/ArticleMetadataSubscriber.php
```diff
@@ -0,0 +1,70 @@
...
+
+    /**
+     * @param ArticleEvent $event
+     */
+    public function populateMetadata(ArticleEvent $event)
+    {
+        $package = $this->getPackage($event);
+        $article = $event->getArticle();
+
+        $article->setMetadata([
```
djbrd-sourcefabric: If all the data is in the package, and persisted, wouldn't it make more sense to keep a reference in the article to the package, and retrieve the data from the package, instead of copying it? I realise that doctrine won't support relations between entities which aren't persisted in the same way, but even so, copying seems like a pretty inelegant solution.

src/SWP/Bundle/ContentBundle/Resources/config/doctrine-phpcr/Article.phpcr.yml
djbrd-sourcefabric: OK, I just noticed that the variable is redeclared in the derived class (which incidentally has the same name as its base class) with a different type. That's pretty funky, I think. You can do it in php, I guess, but whether it's a good idea or not, I wonder.

src/SWP/Bundle/ContentBundle/EventListener/ArticleMetadataSubscriber.php
```diff
@@ -0,0 +1,70 @@
+<?php
+
```
djbrd-sourcefabric: Why set the metadata like this, and not in ArticleFactory::CreateFromPackage()?

src/SWP/Bundle/ContentBundle/Factory/ArticleFactory.php
```diff
@@ -74,8 +76,13 @@ public function create()
      */
     public function createFromPackage(PackageInterface $package)
     {
+        /** @var ArticleInterface $article */
         $article = $this->create();
-        $article->setParentDocument($this->articleProvider->getParent($this->contentRelativePath));
+
+        if ($article instanceof HierarchyInterface) {
```
djbrd-sourcefabric: Are we going to try to support both orm and phpcr? I think we should decide one way or another. Personally, I am against it, but if we are going to do it we should probably do it properly, with tests everywhere, etc. Otherwise any code to support this will be redundant, as no one will use it.

src/SWP/Bundle/ContentBundle/EventListener/ArticleMetadataSubscriber.php
takeit: why would you need to make an extra query if you already have a Package object available for you directly from which you can take all needed data.

src/SWP/Bundle/ContentBundle/EventListener/ArticleMetadataSubscriber.php
takeit: I wanted to separate metadata handling. I might change it. I can maybe add a getter in the package class which will return me such array of needed metadata and in ArticleFactory just do `$article->setMetadata($package->getMetadata())`

src/SWP/Bundle/ContentBundle/Factory/ArticleFactory.php
takeit: we should go with PHPCR in this bundle imo too. I will simply extend `ArticleInterface` by `HierarchyInterface` here so that if statement won't be needed.

src/SWP/Bundle/ContentBundle/EventListener/ArticleMetadataSubscriber.php
djbrd-sourcefabric: Database normalization.

takeit: okay, let's assume I do that. I would need to execute an extra query to get a Package object based on that package id(which makes no sense because I already have a Package object available here in this case) and then from Package I extract metadata and set it into Article object. And I assume you mean to get the Package metadata directly so I dont need to store it in Article object but rather get it from Package object based on referenced id, right? Then in my opinion its wrong, because the metadata should be saved directly under Article object as we work on that object internally in SWP and you would mix phpcr and orm. A Package is just a class needed for stuff related with converting content that comes from Superdesk into our internal Article object. 

We could add `$packageId` property in Article class and store a reference to Package id - it could be useful in cases we want to locate Article by Package id if the content from Superdesk requires, for example, to kill a package or do whatever is needed.

djbrd-sourcefabric: Honestly, I have no idea why you're persisting the Package. Do you use it anywhere? When we ingest the Package, we persist all the data we need from it elsewhere. For me, alarm bells ring when I see the same data being stored twice. If it's an optimization, then do it when you know that it's required. A simple select query is hardly going to make much difference. And normalization is a fundamental principle of database design for a good reason, I guess.

takeit: We store it as a copy of the content that came from Superdesk and the Package is independent from Article. It will be used to preview which data came from Superdesk etc. and that's it. That's why we persist a Package.

I don't have anything against database normalization, I just say it's not needed in this case.

Even if we could get rid off of persisting Package we would need to store the metadata in Article object at the end. So what you describing is not needed at all. To me it looks like over-engineering in this case imo but I might be wrong.

PR #119, opened 2016-08-29
enabled creation and application of rules to map routes to articles from ingested packages

djbrd-sourcefabric:
| Bug fix?                  | [no]
| New feature?              | [yes]
| BC breaks?                | [no]
| Deprecations?             | [no]
| Tests pass?               | [yes]
| Changelog update required | [no]
| Fixed tickets             | [SWP-252]

Rules can now be created to map articles to particular routes, or even to particular templates. These may be assigned priorities in order to ensure which rule takes precedence. The rules are to be written in Symfony's expression language.
TODO: Documentation and validation that submitted rules are valid in terms of syntax


app/config/config_test.yml
```diff
@@ -10,13 +10,13 @@ framework:
...
         collect: false
 
-doctrine:
-    dbal:
-        default_connection: default
-        connections:
-            default:
-                driver:   pdo_sqlite
-                path:     "%kernel.cache_dir%/test.db3"
+#doctrine:
```
ahilles107: why you use real database for tests?

src/SWP/Bundle/ContentBundle/Controller/RouteToArticleController.php
```diff
@@ -0,0 +1,166 @@
...
+     * @ApiDoc(
+     *     resource=true,
+     *     description="Create new RouteToArticle Entity",
+     *     statusCodes={
+     *         201="Returned on success.",
+     *         400="Returned when form has errors"
+     *     },
+     *     input="SWP\Bundle\ContentBundle\Form\RouteToArticleType"
+     * )
+     * @Route("/api/{version}/content/routetoarticle", options={"expose"=true}, defaults={"version"="v1"}, name="swp_api_content_create_routetoarticle")
```
ahilles107: I would change url for all actions in controller to ```/api/{version}/mapping/routes``` Somthing like that - to don't mix with content. As this will be not so often used feature.

src/SWP/Bundle/ContentBundle/Tests/Controller/RouteToArticleControllerTest.php
```diff
@@ -0,0 +1,152 @@
...
+        $this->assertEquals(201, $client->getResponse()->getStatusCode());
+        $this->checkResponsePayload($client, $sent);
+    }
+
+    public function testUpdateRouteToArticleApi()
+    {
+        $sent = [
+            'rule' => 'article.locale matches "/jest/"',
+            'priority' => 1,
+            'routeId' => LoadRoutesData::TEST_NO_CACHE_ROUTE_NAME,
```
ahilles107: this constraint is missing

src/SWP/Bundle/ContentBundle/Model/RouteToArticle.php
```diff
@@ -0,0 +1,166 @@
...
+ * @license http://www.superdesk.org/license
+ */
+namespace SWP\Bundle\ContentBundle\Model;
+
+use SWP\Component\MultiTenancy\Model\TenantAwareInterface;
+
+/**
+ * RouteToArticle.
+ */
+class RouteToArticle implements TenantAwareInterface
```
takeit: this class should also implement `RouteToArticleInterface`

src/SWP/Bundle/ContentBundle/Model/RouteToArticleInterface.php
```diff
@@ -0,0 +1,67 @@
...
+ * AUTHORS and LICENSE files distributed with this source code.
+ *
+ * @copyright 2016 Sourcefabric z.ú.
+ * @license http://www.superdesk.org/license
+ */
+namespace SWP\Bundle\ContentBundle\Model;
+
+use SWP\Component\Storage\Model\PersistableInterface;
+
+interface RouteToArticleInterface extends PersistableInterface
```
takeit: not all methods are defined here, like priority etc

src/SWP/Bundle/ContentBundle/Resources/config/providers.yml
```diff
@@ -17,3 +17,4 @@ services:
             - '@swp.repository.article'
             - '@swp_multi_tenancy.path_builder'
             - '@swp.object_manager.article'
+            - '@swp_content_bundle.routetoarticle.mapper'
```
takeit: `swp_content_bundle.route_to_article.mapper` will be more readable

src/SWP/Bundle/ContentBundle/Service/RouteToArticleMapper.php
```diff
@@ -0,0 +1,90 @@
...
+     * @var ExpressionLanguage
+     */
+    private $language;
+
+    public function __construct(EntityRepository $routeToArticleRepository, DocumentRepository $routeRepository, LoggerInterface $logger)
+    {
+        $this->routeToArticleRepository = $routeToArticleRepository;
+        $this->routeRepository = $routeRepository;
+        $this->logger = $logger;
+        $this->language = new ExpressionLanguage();
```
takeit: You can create a new class `RuleMatcher` and put there logic for matching the rules, register it as a service and inject it here. What if in the future we will want to change the expression language component to something else? You should not change this class code but rather add new class with your own matcher.

app/config/config_test.yml
djbrd-sourcefabric: Just for testing. I shouldn't have checked it in. Will revert this.

src/SWP/Bundle/ContentBundle/Service/RouteToArticleMapper.php
djbrd-sourcefabric: This seems like over-engineering to me. If we really want to decouple this feature from expression language, we will also have to make it possible to override the form used to create and update a RouteToArticleEntity. I really think we should concentrate of keeping things as simple as possible in the short term. We can make the code more flexible later, if and when it needs to be. I think we are more likely to discourage possible contributors by overengineering than to encourage them by making it possible override every feature of the code.

src/SWP/Bundle/ContentBundle/Model/RouteToArticleInterface.php
djbrd-sourcefabric: I've added get and set rule. I assume we don't want get and set tenant code in here.

src/SWP/Bundle/ContentBundle/Service/RouteToArticleMapper.php
ahilles107: ```We can make the code more flexible later``` - and here you get technical debt. Later you can implement other feature not make code more flexible. As this is a service then would be good to change that.

djbrd-sourcefabric: "...if and when it needs to be." In my opinion, it will never need to be. But OK, I will create a service  with 2 lines of code.

src/SWP/Bundle/ContentBundle/Service/RouteToArticleMapper.php
```diff
@@ -0,0 +1,111 @@
...
+ */
+namespace SWP\Bundle\ContentBundle\Service;
+
+use Doctrine\ODM\PHPCR\DocumentRepository;
+use Doctrine\ORM\EntityRepository;
+use Psr\Log\LoggerInterface;
+use SWP\Bundle\ContentBundle\Model\ArticleInterface;
+use SWP\Bundle\ContentBundle\Model\RouteToArticle;
+
+class RouteToArticleMapper
```
takeit: missing interface

src/SWP/Bundle/ContentBundle/Factory/ArticleFactory.php
```diff
@@ -53,11 +59,13 @@ public function __construct(
         FactoryInterface $baseFactory,
         RouteProviderInterface $routeProvider,
         ArticleProviderInterface $articleProvider,
+        RouteToArticleMapper $routeToArticleMapper,
```
takeit: an interface should be used instead, because you are coupling the code strictly with `RouteToArticleMapper` class

src/SWP/Bundle/ContentBundle/Form/Type/RouteToArticleType.php
```diff
@@ -0,0 +1,64 @@
...
+    public function buildForm(FormBuilderInterface $builder, array $options)
+    {
+        $builder
+            ->add('rule', TextType::class, [
+                'required' => false,
+                'constraints' => [
+                    new ArticleRule(),
+                ],
+            ])
+            ->add('priority')
```
takeit: `IntegerType` should be used

src/SWP/Bundle/ContentBundle/Model/RouteToArticle.php
```diff
@@ -0,0 +1,166 @@
...
+     * @var string
+     */
+    protected $tenantCode;
+
+    /**
+     * Get id.
+     *
+     * @return int
+     */
+    public function getId()
```
takeit: `{@inheritdoc}` should be used in getters and setters

src/SWP/Bundle/ContentBundle/Resources/config/services.yml
```diff
@@ -107,3 +108,27 @@ services:
...
             - { name: router, priority: 300 }
+
+    swp_content_bundle.valid_route_id:
+        class: SWP\Bundle\ContentBundle\Validator\Constraints\RouteIdValidator
+        arguments:
+            - '@swp.provider.route'
+        tags:
+            - { name: validator.constraint_validator, alias: route_in_database }
+
+    swp_content_bundle.routetoarticle.repository:
```
takeit: why not using storage bundle to register all needed resources? you define the configuration class, register all services, repositories, factories, object managers by calling one method and you will have a simple configuration, where you will be able to change repository, model class if needed etc., you wont have to register any services in yaml config and you wont have to modify the code.

src/SWP/Bundle/ContentBundle/Form/Type/RouteToArticleType.php
djbrd-sourcefabric: As far as I understand it, this is the default, if priority is defined as an integer in the ORM mapping

src/SWP/Bundle/ContentBundle/Resources/config/services.yml
djbrd-sourcefabric: Well, I'm not the first person to have configured stuff in this way. I just followed the pattern that already appears to be there in that file. If we have standard ways of doing things, maybe these should be documented somewhere. At the moment, I can see quite a number of inconsistencies.

takeit: It is documented in Web Publisher documentation.

ahilles107: http://superdesk-web-publisher.readthedocs.io/en/latest/bundles/SWPStorageBundle/usage.html#register-configured-classes-in-your-extension-class

djbrd-sourcefabric: I see. So you just want me to change the stuff that I've done, or everything that is in that configuration file?

djbrd-sourcefabric: OK, this is changed in the latest commit. I must say that after making the change, I can't really understand the advantage of it. The configuration seemed simpler before and could still be overridden. Anyway, there it is.

ahilles107:
👍  i will merge it after #122 so scrutinizer will run code coverage tests.

takeit:
closed in favour of https://github.com/superdesk/web-publisher/pull/138

PR #120, opened 2016-08-29
[ContentBundle] fixed parent routes assign, routes of collection type are not accessible

takeit:
| Fixed tickets             | [SWP-256](https://dev.sourcefabric.org/browse/SWP-256), [SWP-258](https://dev.sourcefabric.org/browse/SWP-258)

- If there is route of collection type and it has template assigned and has no content assigned, it will load template file assigned to the route
- if route of type of collection has content assigned and this content has template assigned, it will load content's template instead of route template.
- If there is route of collection type and it doesnt have template assigned and has no content assigned, it will return exception that the template is not loaded
- If there is route of collection type and it doesn't have template assigned and has content assigned, it will load template file which is assigned to content
- Additionally if route of collection type is assigned to some articles it is possible to access articles via '/news/<article-slug>' url.

TODO:
- [x] tests
- [ ] docs - separate ticket: [SWP-260](https://dev.sourcefabric.org/browse/SWP-260) - will be done later
- [x] changelog


src/SWP/Bundle/CoreBundle/Controller/ContentController.php
```diff
@@ -25,16 +24,11 @@ class ContentController extends Controller
     /**
      * @param Request $request
      * @param string  $contentTemplate
-     * @param null    $contentDocument
      *
      * @return \Symfony\Component\HttpFoundation\Response
      */
-    public function renderPageAction(Request $request, $contentTemplate, $contentDocument = null)
+    public function renderPageAction(Request $request, $contentTemplate)
```
ahilles107: The parameter $request is not used and could be removed.

PR #123, opened 2016-08-31
[ContentBundle][Bridge] Article's body is not pre-filled

takeit:
| Fixed tickets             | [SWP-263](https://dev.sourcefabric.org/browse/SWP-263)

### DIFF
--- a/src/SWP/Bundle/ContentBundle/Factory/ArticleFactory.php
+++ b/src/SWP/Bundle/ContentBundle/Factory/ArticleFactory.php
@@ -87,7 +87,7 @@ class ArticleFactory implements ArticleFactoryInterface
     {
         /** @var ArticleInterface $article */
         $article = $this->create();
-        
+
         $article->setBody($this->populateBody($package));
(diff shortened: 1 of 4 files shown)


src/SWP/Bundle/ContentBundle/Factory/ArticleFactory.php
```diff
@@ -43,6 +43,16 @@ class ArticleFactory implements ArticleFactoryInterface
     private $contentRelativePath;
 
     /**
+     * @var array
+     */
+    private $allowedTypes = [
```
ahilles107: maybe const?
```
const ALLOWED_TYPES = [
    ItemInterface::TYPE_PICTURE,
    ItemInterface::TYPE_FILE,
    ItemInterface::TYPE_TEXT,
    ItemInterface::TYPE_COMPOSITE,
];
```

takeit: do we want to use it somewhere else? I think its only needed inside this class?

PR #125, opened 2016-09-01
Create fixtures for use on production server

djbrd-sourcefabric:
| Q | A |
| --- | --- |
| Bug fix? | no |
| New feature? | no |
| BC breaks? | no |
| Deprecations? | no |
| Tests pass? | yes |
| Changelog update required | no |
| Fixed tickets | SWP-246 |
| License | AGPLv3 |

Actually, the same fixtures are to be used in both the dev and prod environments. And because it is standard practice to include the fixtures bundles in the prod environment, the commands should be run on a production server in the dev environment (this is the default).


src/SWP/Bundle/FixturesBundle/Resources/fixtures/PHPCR/dev/route.yml
```diff
@@ -0,0 +1,5 @@
+SWP\Bundle\ContentBundle\Doctrine\ODM\PHPCR\Route:
+    route{politics, busines, scitech, health, entertainment, sports, dialogue}:
```
ahilles107: i think that those should be collection not content.

src/SWP/Bundle/FixturesBundle/DataFixtures/PHPCR/LoadArticlesMediaData.php
```diff
@@ -59,6 +59,14 @@ public function loadArticles($env, $manager)
                     'locale' => 'en',
                 ],
             ],
+            'dev' => [
+                [
+                    'title' => 'Test news article',
```
ahilles107: We need more articles on dev/demo - and with more realistic titles (not like this). At least 3-4 articles per route.

src/SWP/Bundle/FixturesBundle/DataFixtures/PHPCR/LoadArticlesData.php
```diff
@@ -199,56 +163,64 @@ public function setRoutesContent($env, $manager)
...
-                    'providers' => [$this],
-                ]
-            );
+        $faker = Factory::create();
+        if ('dev' === $env) {
+            foreach (LoadRoutesData::DEV_ROUTES as $devRouteName) {
+                for ($i = 0; $i < 3; ++$i) {
+                    $articles['dev'][] =
+                        [
+                            'title' => $faker->text(20),
```
ahilles107: catchPhrase would better

PR #153, opened 2016-09-28
[ContentBundle] [TemplatesSystem] Bump php minimum version to ^7.0, simplify ArticlesLoader

ahilles107:
| Fixed tickets             | [SWP-292](https://dev.sourcefabric.org/browse/SWP-292)


.travis.yml
```diff
@@ -33,7 +32,7 @@ before_script:
   - if [[ $TRAVIS_PHP_VERSION != hhvm ]] ; then cat ~/.phpenv/versions/$(phpenv version-name)/etc/php.ini|grep redis; fi;
   - if [[ $TRAVIS_PHP_VERSION != hhvm ]]; then phpenv config-rm xdebug.ini; fi;
   - phpenv rehash
-  - composer install --no-interaction --prefer-dist --dev
+  - composer install --no-interaction --prefer-dist --dev --ignore-platform-reqs
```
takeit: what's the reason to add this?

src/SWP/Bundle/ContentBundle/Provider/ArticleProvider.php
```diff
@@ -65,4 +68,9 @@ public function getParent($id)
     {
         return $this->articleRepository->find($this->pathBuilder->build($id));
     }
+
+    public function getRouteArticlesQuery(string $routeIdentifier, array $order) : SqlQuery
```
takeit: this method should be forced by `ArticleProviderInterface`

src/SWP/Bundle/ContentBundle/Doctrine/ODM/PHPCR/ArticleRepository.php
```diff
@@ -34,4 +38,26 @@ public function findAllArticles()
     {
         return $this->createQueryBuilder('o')->getQuery();
     }
+
+    public function getQueryForRouteArticles(string $identifier, array $order) : SqlQuery
```
takeit: this method should be forced by `ArticleRepositoryInterface`

.travis.yml
ahilles107: it's required for running tests on hhvm (they are not ready yet, but when they will be then tests will try to go again there).

src/SWP/Bundle/ContentBundle/Provider/ArticleProvider.php
ahilles107: done

src/SWP/Bundle/ContentBundle/Doctrine/ODM/PHPCR/ArticleRepository.php
ahilles107: done

takeit: what about also changing other functions signatures to PHP 7 syntax? because one method is php 7 syntax and other are not, I would make it the same

ahilles107: It's optional feature, we don't need to do this now for every method. Let's take it easy and do it one by one when we will touch that function. Option to return `null` or specified type will be in `7.1` - so we can't use it everywhere for now.

PR #154, opened 2016-09-28
Allow to manage article autopublishing feature by rules

takeit:
| Tests pass?               | yes|no
| Fixed tickets             | [SWP-305](https://dev.sourcefabric.org/browse/SWP-305)

TODO:

- [x] - functional tests
- [x] - fix specs
- [x] - documentation

### DIFF
--- a/src/SWP/Bundle/ContentBundle/Rule/Applicator/ArticleRuleApplicator.php
+++ b/src/SWP/Bundle/ContentBundle/Rule/Applicator/ArticleRuleApplicator.php
@@ -37,7 +37,7 @@ final class ArticleRuleApplicator implements RuleApplicatorInterface
     /**
      * @var array
      */
-    private $supportedKeys = ['route', 'templateName'];
+    private $supportedKeys = ['route', 'templateName', 'published'];
 
     /**
      * ArticleRuleApplicator constructor.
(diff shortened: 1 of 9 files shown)


src/SWP/Bundle/ContentBundle/Rule/Applicator/ArticleRuleApplicator.php
```diff
@@ -77,6 +88,10 @@ public function apply(RuleInterface $rule, RuleSubjectInterface $subject)
 
         $subject->setTemplateName($configuration[$this->supportedKeys[1]]);
 
+        if ((bool) $configuration[$this->supportedKeys[2]]) {
```
ahilles107: I don't like relying on array index. Why not `if (in_array('published', $configuration) && true === $configuration['published'])`

takeit: there is option resolver which resolves the configuration array before it goes here and it makes sure that this key is always here.

ahilles107: yeah, but when you add new element at start of array then it's hard to catch that bug.

ahilles107: constant sound better there for me than keys in array ;)

takeit: tests will fail and tell you whats wrong anyway, but yeah I agree referring by key is less readable here tho, will change that

PR #174, opened 2016-11-04
[Templates System Bundle] make bundle and tests separate

ahilles107:
| New feature?             | no
| Fixed tickets             | [SWP-418](https://dev.sourcefabric.org/browse/SWP-418)


src/SWP/Bundle/TemplatesSystemBundle/Resources/config/doctrine/Container.orm.yml
```diff
@@ -42,15 +42,15 @@ SWP\Bundle\TemplateEngineBundle\Model\Container:
             gedmo:
                 timestampable:
                     on: update
-        tenantCode:
-            type: string
-    uniqueConstraints:
-        swp_name_idx:
-          columns: [ name, tenant_code ]
+#        tenantCode:
```
takeit: so currently containers are not tenant aware? If so this `Container` class should be extended by Container class in Core Bundle and the `tenantCode` property should be added and the new class should be used by TemplatesSystemBundle instead

src/SWP/Bundle/TemplatesSystemBundle/Resources/config/doctrine/WidgetModel.orm.yml
```diff
@@ -32,9 +32,9 @@ SWP\Bundle\TemplateEngineBundle\Model\WidgetModel:
             gedmo:
                 timestampable:
                     on: update
-        tenantCode:
-            type: string
+#        tenantCode:
```
takeit: the same in case of wigets

src/SWP/Bundle/TemplatesSystemBundle/Tests/Fixtures/ORM/test/tenant.yml
```diff
@@ -0,0 +1,13 @@
...
+        subdomain: "default"
+        themeName: "swp/test-theme"
+        code: "123abc"
+        organization: "@org1"
+    tenant2:
+        name: "<company()>"
+        subdomain: "client1"
+        themeName: "swp/test-theme"
+        code: "456def"
+        organization: "@org2"
```
takeit: no newline at the end of file

src/SWP/Bundle/TemplatesSystemBundle/Factory/ContainerDataFactory.php
```diff
@@ -0,0 +1,41 @@
...
+ * For the full copyright and license information, please see the
+ * AUTHORS and LICENSE files distributed with this source code.
+ *
+ * @copyright 2015 Sourcefabric z.ú
+ * @license http://www.superdesk.org/license
+ */
+
+namespace SWP\Bundle\TemplatesSystemBundle\Factory;
+
+class ContainerDataFactory
```
takeit: missing interface

src/SWP/Bundle/TemplatesSystemBundle/Factory/ContainerFactory.php
```diff
@@ -0,0 +1,43 @@
...
+ *
+ * @copyright 2015 Sourcefabric z.ú
+ * @license http://www.superdesk.org/license
+ */
+
+namespace SWP\Bundle\TemplatesSystemBundle\Factory;
+
+use SWP\Component\Storage\Factory\FactoryInterface;
+
+class ContainerFactory
```
takeit: this factory is not needed you can use the default one

src/SWP/Bundle/CoreBundle/Controller/ContainerController.php
```diff
@@ -53,7 +53,7 @@ public function listAction(Request $request)
         $entityManager = $this->get('doctrine')->getManager();
         $paginator = $this->get('knp_paginator');
         $containers = $paginator->paginate(
-            $entityManager->getRepository('SWP\Bundle\TemplateEngineBundle\Model\Container')->getAll(),
+            $entityManager->getRepository('SWP\Bundle\CoreBundle\Model\Container')->getAll(),
```
takeit: you can get repository for container: `swp.repository.container` and widget `swp.repository.widget_model` from Symfony container and also use repositories to persist new objects instead of using entity manager strictly

src/SWP/Bundle/CoreBundle/Controller/ContainerController.php
```diff
@@ -14,6 +14,7 @@
 
 namespace SWP\Bundle\CoreBundle\Controller;
```
takeit: this controller can stay in TemplatesSystemBundle now since we decoupled it from fos rest bundle

takeit:
there are also some scrutinizer small issues

src/SWP/Bundle/CoreBundle/Controller/ContainerController.php
ahilles107: It still require `CoreBundle` and `NelmioApiBundle` and i would like to do not have those dependencies here.

takeit: but if I understood correctly the CoreBundle dependency will be optional, It will be required only if you would want to make use of Controllers in our project and if someone will want to make use of it he/she will need to implement its own logic or the similar one we have in core bundle to handle controllers responses properly

ahilles107: Then we need to move reponse classes to `swp/common`.

src/SWP/Bundle/TemplatesSystemBundle/Factory/ContainerFactory.php
takeit: default is `SWP\Component\Storage\Factory\Factory` which should be automatically registered. It does exactly the same as this class.

src/SWP/Bundle/CoreBundle/Controller/ContainerController.php
takeit: or maybe we could put it to `StorageBundle`?

PR #181, opened 2016-11-15
[MenuBundle] Menu drag and drop items, improvements

takeit:
| Deprecations?             | yes|no
| Fixed tickets             | SWP-420, SWP-399


src/SWP/Bundle/CoreBundle/Controller/MenuController.php
```diff
@@ -45,16 +43,79 @@ class MenuController extends Controller
...
+    {
+        $menuItem = $this->findOr404($id);
+        $form = $this->createForm(MenuItemMoveNodeType::class, [], ['method' => $request->getMethod()]);
+        $form->handleRequest($request);
+
+        if ($form->isValid()) {
+            $menuItemManager = $this->get('swp_menu.manager.menu_item');
+            $formData = $form->getData();
+
+            if (isset($formData['parentId'])) {
```
takeit: for now we have two form fields in order to move menu item:
- parentId
- afterId

If `parentId` is given in request then the menu item will be moved to a position directly under the `parentId` subtree (as a first child of that parentId).
If `afterId` is given, the menu item will be moved after the menu item defined by `afterId`.

It works fine but maybe this should be improved a bit, for example we could use fields like:
- parent - parent menu item under which you want to place moved menu item
- position - position in above parent's subtree under which the given element should be placed. This will require sortable gedmo extension tho, but the form will have full validation of given data class and will stick more to the model itself, because `parentId` and `afterId` properties don't exist in the model. Just an idea for the improvement.

PR #184, opened 2016-11-18
AMP HTML integration

takeit:
| Fixed tickets             | -

TODO:
- [x] - enable/disable AMP HTML support per tenant
- [x] - allow to define AMP theme per tenant's theme (currently only one AMP theme for all tenants can be used)
- [x] - docs


app/Resources/amp/amp-theme/index.html.twig
```diff
@@ -0,0 +1,60 @@
+{#
+Available variables:
+- object - Object of an class instance which you provided in the bundle config. e.g. AppBundle\Entity\Post
+#}
+<!doctype html>
```
ahilles107: i would add twig blocks around most important pars - so templators can extend that template and change only needed parts.

PR #206, opened 2016-12-27
[Core Bundle] Check authentication token User credentials between user and current ogranization

ahilles107:
| Fixed tickets             | [SWP-513](https://dev.sourcefabric.org/browse/SWP-513)


src/SWP/Bundle/CoreBundle/Security/Authenticator/TokenAuthenticator.php
```diff
@@ -92,6 +92,7 @@ public function getUser($credentials, UserProviderInterface $userProvider)
         $apiKey->extendValidTo();
         $this->apiKeyRepository->flush();
 
+        /** @var User $user */
```
takeit: `UserInterface` instead `User`

PR #209, opened 2017-01-02
[Templates System] Add option to ignore context meta in gimme and gimmelist loaders

ahilles107:
| Fixed tickets             | []

+ add Loader for Renditions

usage:
```
{% gimmelist media from articleMedia with {'article': gimme.article} %}
    {% gimme rendition with { 'name': '16-9', 'fallback': 'original' } %}
        <img src="{{ url(gimme.rendition) }}" style="width: {{ gimme.rendition.width }}px; height: {{ gimme.rendition.height }}px;" />
    {% endgimme %}
{% endgimmelist %}
```


app/config/config.yml
```diff
@@ -110,7 +110,7 @@ fos_http_cache:
         enabled: true
         generate_url_type: 0
     proxy_client:
-        symfony:
+        varnish:
```
takeit: so by default the varnish needs to be installed to run publisher?

src/SWP/Bundle/CoreBundle/Resources/config/app/config.yml
```diff
@@ -48,6 +48,7 @@ jms_serializer:
                 path: "@SWPCoreBundle/Resources/config/serializer/fos-user"
 
 
+
```
takeit: new line not needed here

src/SWP/Component/TemplatesSystem/Gimme/Context/Context.php
```diff
@@ -283,10 +287,59 @@ public function offsetUnset($name)
...
+
+        return $keysId;
+    }
+
+    /**
+     * @param $id
+     *
+     * @return null|true
+     */
+    public function restoreTemporaryUnseted($id)
```
takeit: name should be `restoreTemporaryUnset`

docs/templates_system/article_media.rst
```diff
@@ -50,3 +50,17 @@ Usage:
         {% endif %}
     {% endgimmelist %}
 
+Get selected rendition only:
+
+
+.. code-block:: twig
+
+    {% gimmelist media from articleMedia with {'article': gimme.article} %}
```
ahilles107: Nope, as in publisher we don't have Article renditions. Renditions are per article media (type image). This will can be used like that when we will add support for Superdesk featured image. It will can look like that then:

```
{% gimme rendition with { 'name': '16-9', 'source': 'featuredImage' } %}
    <img src="{{ url(gimme.rendition) }}" style="width: {{ gimme.rendition.width }}px; height: {{ gimme.rendition.height }}px;" />
{% endgimme %}
```

Then we could check for featured image in context current article.

app/config/config.yml
ahilles107: nope, it's not needed - just invalidation will not work.

src/SWP/Component/TemplatesSystem/Gimme/Context/Context.php
ahilles107: done

PR #219, opened 2017-01-17
Automatically create menu widget when root menu/navigation is created

takeit:
| Fixed tickets             | SWP-563

TODO:
- [x] - functional tests
- [x] - docs
- [x] - remove widget when menu is removed

### DIFF
--- a/src/SWP/Bundle/MenuBundle/MenuEvents.php
+++ b/src/SWP/Bundle/MenuBundle/MenuEvents.php
@@ -1,0 +1,42 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of the Superdesk Web Publisher Menu Bundle.
+ *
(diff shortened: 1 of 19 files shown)


src/SWP/Bundle/MenuBundle/MenuEvents.php
```diff
@@ -0,0 +1,42 @@
...
+    /**
+     * The MENU_DELETED event occurs after menu is deleted.
+     *
+     * @Event("Symfony\Component\EventDispatcher\GenericEvent")
+     *
+     * @var string
+     */
+    const MENU_DELETED = 'swp.menu.deleted';
+
+    private function __construct()
```
ahilles107: why You need that empty constructor?

takeit: so the class won't be instantiated as its not needed to do it, it just keeps events as consts. Note the construct is private.

PR #225, opened 2017-01-25
[Maintenance] add option to configure trusted proxies from parametersyml

ahilles107:
| Fixed tickets             | [SWP-556](https://dev.sourcefabric.org/browse/SWP-556)

### DIFF
--- a/app/config/parameters.yml.dist
+++ b/app/config/parameters.yml.dist
@@ -38,3 +38,8 @@ parameters:
     # Array of authorized superdesk servers
     superdesk_servers:
         - superdesk.dev
+
+    # Array of trusted proxies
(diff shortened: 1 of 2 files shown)


app/config/parameters.yml.dist
```diff
@@ -38,3 +38,8 @@ parameters:
     # Array of authorized superdesk servers
     superdesk_servers:
         - superdesk.dev
+
+    # Array of trusted proxies
+    trusted_proxies:
+        - 192.0.0.1
+        - 10.0.0.0/8
```
takeit: maybe let's make it `~` by default? Because it will not work if the IP of the reverse proxy address is changing all the time

PR #235, opened 2017-01-30
[Core Bundle] Add api for rendering single container (and it content)

ahilles107:
| Fixed tickets             | [SWP-572](https://dev.sourcefabric.org/browse/SWP-572)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ in 0.1 minor versions.
 To get the diff for a specific change, go to https://github.com/superdesk/web-publisher/commit/XXX where XXX is the change hash
 
 * 0.2.0
+ * feature [#235] Add API endpoint for rendering single container (and its content)
  * feature [#219] Automatically create menu widget when root menu/navigation is created
  * feature [#218] Assign article to route of type content automatically when article is published
  * feature [#215] Filter articles by metadata in gimmelist
--- a/src/SWP/Bundle/CoreBundle/Controller/ContainerController.php
+++ b/src/SWP/Bundle/CoreBundle/Controller/ContainerController.php
@@ -14,6 +14,7 @@
 
 namespace SWP\Bundle\CoreBundle\Controller;
 
+use SWP\Component\TemplatesSystem\Gimme\Model\ContainerInterface;
 use SWP\Component\TemplatesSystem\Gimme\Model\WidgetModelInterface;
 use Symfony\Bundle\FrameworkBundle\Controller\Controller;
 use Nelmio\ApiDocBundle\Annotation\ApiDoc;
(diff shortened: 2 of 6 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ in 0.1 minor versions.
 To get the diff for a specific change, go to https://github.com/superdesk/web-publisher/commit/XXX where XXX is the change hash
 
 * 0.2.0
+ * feature [#235] Add API endpoint for rendering single container (and his content)
```
takeit: `and its content`

src/SWP/Bundle/CoreBundle/Controller/ContainerController.php
```diff
@@ -95,6 +96,39 @@ public function getAction(Request $request, $id)
...
+     *     }
+     * )
+     * @Route("/api/{version}/templates/containers/{id}/render", requirements={"id"="\d+"}, options={"expose"=true}, defaults={"version"="v1"}, name="swp_api_templates_render_container")
+     * @Method("GET")
+     * @Cache(expires="10 minutes", public=true)
+     */
+    public function renderAction($id)
+    {
+        /** @var ContainerInterface $container */
+        $container = $this->get('doctrine')->getManager()
```
takeit: why `doctrine` service and not `swp.repository.container`?

src/SWP/Bundle/CoreBundle/Controller/ContainerController.php
```diff
@@ -95,6 +96,39 @@ public function getAction(Request $request, $id)
...
+     *
+     * @ApiDoc(
+     *     resource=true,
+     *     description="Render single container and it's widgets.",
+     *     statusCodes={
+     *         200="Returned on success.",
+     *         404="Container not found"
+     *     }
+     * )
+     * @Route("/api/{version}/templates/containers/{id}/render", requirements={"id"="\d+"}, options={"expose"=true}, defaults={"version"="v1"}, name="swp_api_templates_render_container")
```
takeit: what about returning the content based on the content negotiation? If we allow to return not only json but also html, content negotiation is required in this case. I mean that this endpoint is not needed then, it can be done using the existing one but return a proper data in a proper format if requested, if the content type is not supported by the endpoint we should return a proper status code with the error. The api is inconsistent in that way because it only accepts application/json but can return also text/html. We should handle it properly imo.

ahilles107: 👍  - i will make a ticket for this feature

takeit: okay, it is just a matter of adding a config to fosrestbundle afaik to support this, but since we don't extend fos rest controller it will need some more work I guess

ahilles107: And i will change output format here to json with content as value of `content` property.

PR #246, opened 2017-02-07
[ContentBundle] add support for FeatureMedia, add option to provide selected media on rendition loader

ahilles107:
| Fixed tickets             | [SWP-541](https://dev.sourcefabric.org/browse/SWP-541)


app/migrations/Version20170207132939.php
```diff
@@ -0,0 +1,39 @@
...
+
+    /**
+     * @param Schema $schema
+     */
+    public function down(Schema $schema)
+    {
+        // this down() migration is auto-generated, please modify it to your needs
+        $this->abortIf($this->connection->getDatabasePlatform()->getName() !== 'postgresql', 'Migration can only be executed safely on \'postgresql\'.');
+
+        $this->addSql('CREATE SCHEMA public');
```
takeit: won't that fail here if you revert that migration because schema already exists? At least it was failing in my case but pls check

PR #249, opened 2017-02-09
Swp 588 render custom class with widgets

ahilles107:
| Fixed tickets             | [SWP-588](https://dev.sourcefabric.org/browse/SWP-588)

### DIFF
--- a/src/SWP/Bundle/TemplatesSystemBundle/Container/SimpleContainer.php
+++ b/src/SWP/Bundle/TemplatesSystemBundle/Container/SimpleContainer.php
@@ -15,6 +15,7 @@
 namespace SWP\Bundle\TemplatesSystemBundle\Container;
 
 use SWP\Component\TemplatesSystem\Gimme\Model\ContainerInterface;
+use SWP\Component\TemplatesSystem\Gimme\Widget\WidgetHandlerInterface;
 
 class SimpleContainer
 {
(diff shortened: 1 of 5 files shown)


src/SWP/Bundle/TemplatesSystemBundle/Container/SimpleContainer.php
```diff
@@ -98,7 +100,7 @@ public function renderWidgets()
     {
         $widgetsOutput = [];
         foreach ($this->widgets as $widget) {
-            $widgetsOutput[] = $widget->render();
+            $widgetsOutput[] = '<div class="'.self::WIDGET_CLASS.'">'.$widget->render().'</div>';
```
takeit: it's not only a class but also the id should be added, see [SWP-604](https://dev.sourcefabric.org/browse/SWP-604) - I think those tickets were partially duplicated.

src/SWP/Bundle/TemplatesSystemBundle/Container/SimpleContainer.php
```diff
@@ -97,8 +100,9 @@ public function hasWidgets()
     public function renderWidgets()
     {
         $widgetsOutput = [];
+        /** @var WidgetHandlerInterface $widget */
         foreach ($this->widgets as $widget) {
-            $widgetsOutput[] = $widget->render();
+            $widgetsOutput[] = '<div id="'.self::WIDGET_CLASS.'_'.$widget->getId().'" class="'.self::WIDGET_CLASS.'">'.$widget->render().'</div>';
```
takeit: I would use `sprintf` function here

PR #270, opened 2017-02-22
Rely on evelovedfrom property instead of a slug

takeit:
| Fixed tickets             | SWP-589


app/migrations/Version20170222142920.php
```diff
@@ -0,0 +1,40 @@
...
+     */
+    public function up(Schema $schema)
+    {
+        // this up() migration is auto-generated, please modify it to your needs
+        $this->abortIf($this->connection->getDatabasePlatform()->getName() !== 'postgresql', 'Migration can only be executed safely on \'postgresql\'.');
+
+        $this->addSql('ALTER TABLE swp_item ADD evolved_from VARCHAR(255) DEFAULT NULL');
+        $this->addSql('ALTER TABLE swp_package ADD evolved_from VARCHAR(255) DEFAULT NULL');
+        $this->addSql('ALTER TABLE swp_article ADD code VARCHAR(255) DEFAULT NULL');
+        $this->addSql('UPDATE swp_article SET code = md5(random()::text)');
```
ahilles107: It creates issues, after that you can't update article as is no longer recognized and Publisher want's to create new one with this same slug (and it throws exception). As in real article createion we use guuid here initially then also here we should fill it with guid.

PR #282, opened 2017-03-02
Resolve tenant by domain name

ahilles107:
| Fixed tickets             | not yet

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -94,7 +94,8 @@
         "facebook/facebook-instant-articles-sdk-php": "v1.5.*",
         "takeit/amp-html-bundle": "^0.2.0@dev",
         "asm89/twig-cache-extension": "^1.3",
-        "doctrine/doctrine-migrations-bundle": "^1.0"
+        "doctrine/doctrine-migrations-bundle": "^1.0",
+        "layershifter/tld-extract": "^1.2"
     },
     "require-dev": {
         "phpspec/phpspec": "^2.4",
--- a/src/SWP/Bundle/MultiTenancyBundle/Resources/config/services.yml
+++ b/src/SWP/Bundle/MultiTenancyBundle/Resources/config/services.yml
@@ -17,7 +17,6 @@ services:
     swp_multi_tenancy.tenant_resolver:
         class: '%swp_multi_tenancy.tenant_resolver.class%'
         arguments:
-            - '%domain%'
             - '@swp.repository.tenant'
 
     swp_multi_tenancy.tenant_context:
--- a/src/SWP/Component/MultiTenancy/Repository/TenantRepositoryInterface.php
+++ b/src/SWP/Component/MultiTenancy/Repository/TenantRepositoryInterface.php
@@ -28,8 +28,6 @@ interface TenantRepositoryInterface extends RepositoryInterface
      * @param string $subdomain The subdomain
      * @param string $domain    The domain
      *
-     * @throws \Doctrine\ORM\NonUniqueResultException
-     *
      * @return TenantInterface|null The instance of TenantInterface or null
      */
     public function findOneBySubdomainAndDomain($subdomain, $domain);
@@ -39,8 +37,6 @@ interface TenantRepositoryInterface extends RepositoryInterface
      *
      * @param string $domain The domain
      *
-     * @throws \Doctrine\ORM\NonUniqueResultException
-     *
      * @return TenantInterface|null The instance of TenantInterface or null
      */
     public function findOneByDomain($domain);
@@ -49,8 +45,6 @@ interface TenantRepositoryInterface extends RepositoryInterface
      * Finds the tenant by code.
      *
      * @param string $code The code
-     *
-     * @throws \Doctrine\ORM\NonUniqueResultException
      *
      * @return TenantInterface|null The instance of TenantInterface or null
      */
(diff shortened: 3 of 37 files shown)


composer.json
```diff
@@ -94,7 +94,8 @@
         "facebook/facebook-instant-articles-sdk-php": "v1.5.*",
         "takeit/amp-html-bundle": "^0.2.0@dev",
         "asm89/twig-cache-extension": "^1.3",
-        "doctrine/doctrine-migrations-bundle": "^1.0"
+        "doctrine/doctrine-migrations-bundle": "^1.0",
+        "layershifter/tld-extract": "^1.2"
```
takeit: add this also to composer.json in  multitenancy component or is this lib auctually needed needed for this simple use case? you can easily extract domain using own php implementation using explode

src/SWP/Bundle/MultiTenancyBundle/Resources/config/services.yml
```diff
@@ -17,7 +17,6 @@ services:
     swp_multi_tenancy.tenant_resolver:
         class: '%swp_multi_tenancy.tenant_resolver.class%'
         arguments:
-            - '%domain%'
```
takeit: if it is not used anymore it should be removed also from the parameters.yml.dist

src/SWP/Component/MultiTenancy/Resolver/TenantResolver.php
```diff
@@ -49,21 +42,36 @@ public function __construct($domain, TenantRepositoryInterface $tenantRepository
      */
     public function resolve($host = null)
     {
-        if (null === $host) {
-            $host = self::DEFAULT_TENANT;
```
takeit: that const should be removed from TenantResolverInterface because it is not used anywhere now afaik

src/SWP/Component/MultiTenancy/Repository/TenantRepositoryInterface.php
```diff
@@ -26,12 +26,24 @@
...
      */
-    public function findOneBySubdomain($subdomain);
+    public function findOneBySubdomainAndDomain($subdomain, $domain);
+
+    /**
+     * Finds the tenant by domain.
+     *
+     * @param string $domain The domain
+     *
+     * @throws \Doctrine\ORM\NonUniqueResultException
```
takeit: by this u are making a dependency on Doctrine ORM which is not included into that component and it shouldnt. This interface is generic, it shouldnt rely on doctrine only, it shouldnt tell you it throws doctrine orm exception

ahilles107: Ok ;) But You added it originally - https://github.com/superdesk/web-publisher/commit/cc17be0f80f16d654f0458a73cc6d2ce86f4d955#diff-6d78e0c57d88d61b951b127912ddc261R30

takeit: then it should be removed :)

src/SWP/Bundle/MultiTenancyBundle/Resources/config/services.yml
ahilles107: it is used for migration (if there is no domain then it's filled with this). We can depreciate it but should stay until next minor.

src/SWP/Component/MultiTenancy/Resolver/TenantResolver.php
```diff
@@ -49,21 +42,50 @@ public function __construct($domain, TenantRepositoryInterface $tenantRepository
...
     }
 
     /**
+     * @param $host
+     *
+     * @return string
+     */
+    protected function extractDomain($host)
+    {
+        if (null === $host || 'localhost' === $host) {
```
takeit: set `localhost` as const

src/SWP/Component/MultiTenancy/Resolver/TenantResolver.php
```diff
@@ -49,21 +42,50 @@ public function __construct($domain, TenantRepositoryInterface $tenantRepository
...
+            return 'localhost';
+        }
+
+        $extract = new \LayerShifter\TLDExtract\Extract();
+        $result = $extract->parse($host);
+
+        // handle case for ***.localhost
+        if ('localhost' == $result->getSuffix() && null !== $result->getHostname() && null === $result->getSubdomain()) {
```
takeit: why not `===`?

src/SWP/Component/MultiTenancy/Resolver/TenantResolver.php
```diff
@@ -72,16 +94,19 @@ public function resolve($host = null)
      */
     protected function extractSubdomain($host)
     {
-        if ($this->domain === $host) {
-            return self::DEFAULT_TENANT;
+        $extract = new \LayerShifter\TLDExtract\Extract();
```
takeit: you can move that part of code to a provate method because you are calling it in two places

src/SWP/Component/MultiTenancy/Resolver/TenantResolver.php
ahilles107: it should be === it was late and i was hungry :)

PR #291, opened 2017-03-08
SWP-703 - Handle items and packages updates properly & SWP-668 organization's articles improvements

takeit:
| Fixed tickets             | SWP-703, SWP-668

SWP-703
- added an option to update articles

SWP-668
- newly pushed articles are created with status `new` and are not published by default, unless a business rule is set to do so automatically,
- newly pushed articles do not have tenant/website assigned by default, only an organization,
- newly pushed articles are not available via `/content/articles/` API endpoint because they are not allowed to appear there without tenant,
- newly pushed articles are available via `/organization/articles` endpoint, which returns a list of all articles for current organization in the system, even those without tenant/website assigned. Organization is resolved by current tenant/website url,
- newly pushed articles, without tenant/website assigned, can be published only via `/content/articles/<article-id>` endpoint, using `PATCH` method,
- article without tenant/website assigned can be retrieved by calling `/organization/articles/<article-id>` using `GET` method.


src/SWP/Bundle/CoreBundle/Controller/ContentPushController.php
```diff
@@ -0,0 +1,58 @@
...
+     *     resource=true,
+     *     description="Adds a new content from HTTP Push",
+     *     statusCodes={
+     *         201="Returned on successful post."
+     *     }
+     * )
+     * @Route("/api/{version}/content/push", options={"expose"=true}, defaults={"version"="v1"}, name="swp_api_content_push")
+     * @Method("POST")
+     */
+    public function pushContentAction(Request $request)
```
ahilles107: why you override it?

takeit: because I need to use multitenancy code to disable tenantable filter in order to check if an article already exists and I didn't include whole multitenancy to content bundle.

If an article was pushed already there is no tenant defined for it so if you update the same article it would then return `null` when chcking for article existence because tenantable filter will append `tenant_code` which doesn't exist.

PR #300, opened 2017-03-16
Allow to specify widget type by his class instead id

ahilles107:
| Fixed tickets             | SWP-651

### DIFF
--- a/src/SWP/Bundle/CoreBundle/Controller/WidgetController.php
+++ b/src/SWP/Bundle/CoreBundle/Controller/WidgetController.php
@@ -69,7 +69,7 @@ class WidgetController extends FOSRestController
      * @Route("/api/{version}/templates/widgets/{id}", requirements={"id"="\d+"}, options={"expose"=true}, defaults={"version"="v1"}, name="swp_api_templates_get_widget")
      * @Method("GET")
      */
-    public function getAction(Request $request, $id)
+    public function getAction($id)
     {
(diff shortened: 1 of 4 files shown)


src/SWP/Bundle/CoreBundle/Controller/WidgetController.php
```diff
@@ -84,6 +84,11 @@ public function getAction(Request $request, $id)
     /**
      * Create new widget.
      *
+     * Note:
+     *
+     *     Widget Type can be widget ID or his class (string).
+     *     Example: "SWP\\Bundle\\CoreBundle\\Widget\\ContentListWidget"
```
takeit: This fix doesn't change anything because type of the widget will be still a class name `SWP\\Bundle\\CoreBundle\\Widget\\ContentListWidget` which is not readable for end user. A mapping should introduced so for example `SWP\\Bundle\\CoreBundle\\Widget\\ContentListWidget` would map to `list` type which is more user friendly and internally we could use class namespace. So in request to create new form you would type `list` widget type. At least ticket is about it afaik.

ahilles107: Ticket is about this that you provide `integer` and get `string`. So this fix that without breaking BC. Short names for WidgetsHandler is something else and should be discussed. I see issue with storing class as type name (as we do this now) as we will need to run migration if we change class namespace or name. This is fast fix. You talk about change in concept.

PR #336, opened 2017-05-15
ElasticSearch integration & bugfixes

takeit:
| Fixed tickets             | SWP-790, SWP-797, SWP-796


src/SWP/Bundle/ElasticSearchBundle/Controller/Api/PackageSearchController.php
```diff
@@ -42,7 +42,9 @@ class PackageSearchController extends Controller
      *         {"name"="authors", "dataType"="string", "pattern"="John Doe | John Doe, Matt Smith"},
      *         {"name"="term", "dataType"="string", "pattern"="search phrase"},
      *         {"name"="sorting", "dataType"="string", "pattern"="-id|id"},
-     *         {"name"="source", "dataType"="string"}
+     *         {"name"="source", "dataType"="string"},
+     *         {"name"="per_page", "dataType"="integer"},
```
ahilles107: for what is that? It's pagination? Then it should be called `limit`.

takeit: yes, I can change it it doesnt matter really imo, I will add description for each filter in the doc yet

ahilles107: In every api we have `limit` so it should be like this here also.

composer.json
```diff
@@ -96,7 +96,8 @@
         "takeit/amp-html-bundle": "^0.2.0",
         "asm89/twig-cache-extension": "^1.3",
         "doctrine/doctrine-migrations-bundle": "^1.0",
-        "layershifter/tld-extract": "^1.2"
+        "layershifter/tld-extract": "^1.2",
+        "friendsofsymfony/elastica-bundle": "^4.0"
```
ahilles107: it shouldn't be in `ElasticSearchBundle`?

takeit: it depends how we want to keep that elasticsearch integration, whether it should be part of the publisher project, like other Publisher bundles, or as an extension/plugin which should be installed independently as a 3rd party bundle

app/config/parameters.yml.dist
```diff
@@ -45,3 +45,5 @@ parameters:
     trusted_proxies:
         - 192.0.0.1
         - 10.0.0.0/8
+    env(ELASTICA_HOST): localhost
+    env(ELASTICA_PORT): 9200
```
ahilles107: missing new line

docs/manual/templates_system/search.rst
```diff
@@ -0,0 +1,87 @@
...
+
+and the ``search_results.html.twig`` template:
+
+    .. code-block:: twig
+
+        <form name="filter" method="get">
+        <input type="search" id="filter_search" name="q">
+        </form>
+
+        {% for item in results %}
```
ahilles107: `item` is Meta? It should be.

src/SWP/Bundle/ContentBundle/Loader/ArticleMediaLoader.php
```diff
@@ -92,7 +92,7 @@ public function load($type, $parameters = [], $responseType = LoaderInterface::C
             }
 
             $criteria = $this->applyPaginationToCriteria($criteria, $parameters);
-            if ($criteria->get('article')->getMedia()->isInitialized()) {
```
ahilles107: Why you removed it? there is no point to lazy load it here - in else we have faster method for count.

src/SWP/Bundle/ElasticSearchBundle/Controller/Api/ArticleSearchController.php
```diff
@@ -0,0 +1,90 @@
...
+                'publishedAt' => $request->query->get('publishedAt'),
+                'tenantCode' => $currentTenant->getCode(),
+                'sources' => array_filter((array) $request->query->get('source', [])),
+                'metadata' => array_filter((array) $request->query->get('metadata', [])),
+            ]
+        );
+
+        $repositoryManager = $this->get('fos_elastica.manager');
+        /** @var ArticleRepository $repository */
+        $repository = $repositoryManager->getRepository($this->getParameter('swp.model.article.class'));
```
ahilles107: why not get repository from container with `swp.repository.article`?

takeit: because its not Publisher repository but elastic search repository.

PR #342, opened 2017-05-25
[UserBundle] add API to promote and demote users (add/remove user roles)

ahilles107:
| Fixed tickets             | SWP-743

### DIFF
--- a/src/SWP/Bundle/UserBundle/Form/Type/UserRolesType.php
+++ b/src/SWP/Bundle/UserBundle/Form/Type/UserRolesType.php
@@ -1,0 +1,53 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of the Superdesk Web Publisher User Bundle.
+ *
+ * Copyright 2016 Sourcefabric z.ú. and contributors.
+ *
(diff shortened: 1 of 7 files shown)


src/SWP/Bundle/UserBundle/Form/Type/UserRolesType.php
```diff
@@ -0,0 +1,53 @@
...
+
+class UserRolesType extends AbstractType
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function buildForm(FormBuilderInterface $builder, array $options)
+    {
+        $builder
+            ->add('roles', TextType::class)
```
takeit: why not choice type?

ahilles107: With choice we would need hardcode all possible roles, and with this it's more flexible. We could also accept child form type here with option to add many entries but it would be to much complicated. IMO it's simple and does job good.

takeit: we can always use some roles service that will return list of all available roles instead of hardcoding it but we can refactor it later when needed also.

PR #344, opened 2017-05-26
Pass List Id to content list widget

ahilles107:
| Fixed tickets             | SWP-607

DEPRECATION:

From now you should use manual fetching content list items in content list widget templates. Content list loader is much more powerful than iteration on raw list data.

Documentation for content list items loader is here: http://superdesk-publisher.readthedocs.io/en/latest/manual/templates_system/content_list.html

### DIFF
--- a/src/SWP/Bundle/CoreBundle/Widget/ContentListWidget.php
+++ b/src/SWP/Bundle/CoreBundle/Widget/ContentListWidget.php
@@ -18,7 +18,6 @@ namespace SWP\Bundle\CoreBundle\Widget;
 
 use SWP\Bundle\TemplatesSystemBundle\Widget\TemplatingWidgetHandler;
 use SWP\Component\ContentList\Model\ContentListInterface;
-use SWP\Component\ContentList\Repository\ContentListRepositoryInterface;
 
 final class ContentListWidget extends TemplatingWidgetHandler
 {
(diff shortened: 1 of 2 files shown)


src/SWP/Bundle/CoreBundle/Widget/ContentListWidget.php
```diff
@@ -40,13 +39,12 @@ public function render()
...
         $listId = (int) $this->getModelParameter('list_id');
 
-        /** @var ContentListRepositoryInterface $contentListRepository */
-        $contentListRepository = $this->getContainer()->get('swp.repository.content_list');
         /** @var ContentListInterface $contentList */
-        $contentList = $contentListRepository->findListById($listId);
+        $contentList = $this->getContainer()->get('swp.repository.content_list')->findListById($listId);
 
         return $this->renderTemplate($templateName, [
             'contentList' => $contentList,
```
takeit: then I guess we dont need to pass the whole content list here, just the required params (listId, limit). Because you are passing now listId but you can actually get it from contentList.

ahilles107: Yeah, but `contentList` is depreceated. I will remove it soon.  It will stay now to do not break existing templates and sites (give some time for Ljuba to fix it).

PR #346, opened 2017-05-31
set default value for urgency, handle genre as array in package item

ahilles107:
| Fixed tickets             | 


takeit:
what about Ninjs validator schema update for `genre`? does it pass the validation without that field?

PR #356, opened 2017-06-12
don't try to fetch theme if tenant is not set in context

ahilles107:
| Fixed tickets             | 


src/SWP/Bundle/CoreBundle/Theme/TenantAwareThemeContext.php
```diff
@@ -72,6 +72,9 @@ public function getTheme()
     {
         /* @var ThemeAwareTenantInterface $tenant */
         $tenant = $this->tenantContext->getTenant();
+        if (null === $tenant) {
```
takeit: why it worked before?

ahilles107: probably it wasn't called on cache clear. In `cache:clear` command we don't have tenant - we shouldn't have so this is ok.

PR #382, opened 2017-08-09
Swp 920

ljubar:
| Fixed tickets             | SWP-920


src/SWP/Bundle/FixturesBundle/Resources/themes/DefaultTheme/views/base.html.twig
```diff
@@ -87,7 +87,7 @@
         {% endcontainer %}
 
         {# Trending topics is array of first in a row keyword of three latest articles #}
-        {% cache 'subMenu' { time:3600 } %}
+        {% cache 'subMenu' { time:0 } %}
```
ahilles107: if time is 0 then why you need cache block here?

PR #385, opened 2017-08-17
upgrade symfony to 3.3 release, fix depreceations, use twig 2.0

ahilles107:
| Fixed tickets             | none


app/config/config.yml
```diff
@@ -38,7 +38,6 @@ framework:
         engines: ['twig']
     default_locale:  "%locale%"
     trusted_hosts:   ~
-    trusted_proxies: "%trusted_proxies%"
```
takeit: is this setting deprecated? if so you should also remove param from parameters.yml.dist file

PR #395, opened 2017-09-08
add missing migrations

ahilles107:
| Fixed tickets             | SWP-779

### DIFF
--- a/app/migrations/Version20170908083733.php
+++ b/app/migrations/Version20170908083733.php
@@ -1,0 +1,64 @@
+<?php
+
+namespace SWP\Migrations;
+
+use Doctrine\DBAL\Migrations\AbstractMigration;
+use Doctrine\DBAL\Schema\Schema;
+
+/**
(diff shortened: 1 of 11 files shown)


app/migrations/Version20170908083733.php
```diff
@@ -0,0 +1,64 @@
...
+{
+    /**
+     * @param Schema $schema
+     */
+    public function up(Schema $schema)
+    {
+        // this up() migration is auto-generated, please modify it to your needs
+        $this->abortIf($this->connection->getDatabasePlatform()->getName() !== 'postgresql', 'Migration can only be executed safely on \'postgresql\'.');
+
+        $this->addSql('CREATE SEQUENCE swp_event_category_id_seq INCREMENT BY 1 MINVALUE 1 START 1');
```
takeit: all related events tables should not be added, it not used anyhow in publisher

ahilles107: but it's in source code and it's blocking installing demo data

PR #401, opened 2017-10-15
improve env usage, install rabbitmq

ahilles107:
| Fixed tickets             | 


src/SWP/Bundle/AnalyticsBundle/Controller/StatisticsController.php
```diff
@@ -0,0 +1,40 @@
...
+     *
+     * @Method("GET|POST")
+     *
+     * Receive events
+     *
+     * @return \Symfony\Component\HttpFoundation\Response
+     */
+    public function receiveAction()
+    {
+        return new Response('received');
```
takeit: This is missing the functionality, why do you push this controller at all?

app/AppKernel.php
```diff
@@ -54,6 +54,7 @@ public function registerBundles()
 
             new FOS\ElasticaBundle\FOSElasticaBundle(),
             new SWP\Bundle\ElasticSearchBundle\SWPElasticSearchBundle(),
+            new OldSound\RabbitMqBundle\OldSoundRabbitMqBundle(),
```
takeit: Do we use it somewhere? I don't see any usage of this besides the configuration.

ahilles107: we will in next PR, don't want to remove it now.

src/SWP/Bundle/AnalyticsBundle/Controller/StatisticsController.php
ahilles107: i will improve it later this week, don't want to remove it now.

PR #424, opened 2017-12-29
Create routes defined in theme config

ahilles107:
| Fixed tickets             | SWP-938


app/Resources/views/article.html.twig
```diff
@@ -0,0 +1,8 @@
+<html>
+    <head>
+        <title>Superdesk Web Publisher</title>
+    </head>
+    <body>
+        <h3>This is default article.html.twig template. Please replace it by creating your own version in used theme (views/article.html.twig).</h3>
```
takeit: `This is default "article.html.twig" template file. Please, replace it by creating your own version of this file in the currently used theme.` sounds much clearer.`

docs/cookbooks/templates/theme_generated_elements.rst
```diff
@@ -0,0 +1,125 @@
...
+Thanks to this feature theme developer can be sure that all of his article or category templates will be visible without
+complicated configuration by end user.
+
+Generated elements can be declared in theme.json config file under :code:`generatedData` key. Example:
+
+.. code-block:: json
+
+    {
+     "name": "my/custom-theme",
+     "generatedData": { ... }
```
takeit: I would rename `generatedData` to `defaultData` or `demoData`

docs/manual/themes/assets.rst
```diff
@@ -0,0 +1,87 @@
+Work with theme assets
+----------------------
+To install theme assets you need to run ``sylius:theme:assets:install`` command.
+
+Theme assets (JavaScript, CSS etc. files) should be placed inside the theme directory. There are few ways of reading
+theme assets in your Twig templates. The below how-to describes where to place the assets, how to install it and use it.
+
+Load assets from the theme's public directory (``app/themes/<theme-name>/public``)
```
takeit: I would remove `(``app/themes/<theme-name>/public``)` from headline as it will render ugly in ToC.

src/SWP/Bundle/ContentBundle/Hydrator/ArticleHydrator.php
```diff
@@ -80,8 +72,7 @@ public function hydrate(ArticleInterface $article, PackageInterface $package): A
         $article->setLead($this->populateLead($package));
         $article->setMetadata($package->getMetadata());
         $article->setKeywords($package->getKeywords());
-        // assign default route
-        $article->setRoute($this->routeProvider->getRouteForArticle($article));
+        $article->setRoute($article->getRoute());
```
takeit: why do you set already set route? I can be removed IMO.

src/SWP/Bundle/ContentBundle/Model/Route.php
```diff
@@ -76,33 +71,26 @@ public function __construct()
...
 
     /**
-     * {@inheritdoc}
+     * @param ArticleInterface $article
      */
-    public function getRouteName()
+    public function addArticle(ArticleInterface $article): void
     {
-        return $this->getName();
+        $this->articles->add($article);
```
takeit: you can check first here if an article exists, if not then add it, else skip

src/SWP/Bundle/CoreBundle/Command/ThemeSetupCommand.php
```diff
@@ -118,22 +130,31 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+                $theme = $container->get('sylius.context.theme')->getTheme();
+                $requiredDataProcessor = $container->get('swp_core.processor.theme.required_data');
+                $requiredDataProcessor->processTheme($theme);
+                $output->writeln('<info>Theme required data was persisted successfully!</info>');
             }
         } catch (\Exception $e) {
             $output->writeln('<error>Theme could not be installed!</error>');
-            $output->writeln('<error>Stacktrace: '.$e->getMessage().'</error>');
+            $output->writeln('<error>Error message: '.$e->getMessage().'</error>');
+            //$output->writeln('<error>Stack trace: '.$e->getTraceAsString().'</error>');
```
takeit: This comment can be removed

src/SWP/Bundle/CoreBundle/Model/Route.php
```diff
@@ -14,222 +14,18 @@
...
 
 /**
  * Class Route.
  */
-class Route extends BaseRoute implements PersistableInterface, RouteInterface, TenantAwareInterface, ArticlesCountInterface
+class Route extends BaseRoute implements RouteInterface, TenantAwareInterface, ArticlesCountInterface
 {
     use TenantAwareTrait, ArticlesCountTrait;
 
-    /**
```
takeit: why are all properties removed from this class?

src/SWP/Bundle/CoreBundle/Theme/Processor/RequiredDataProcessor.php
```diff
@@ -0,0 +1,79 @@
...
+        $this->themeWidgetsGenerator = $themeWidgetsGenerator;
+        $this->themeContentListsGenerator = $themeContentListsGenerator;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function processTheme(ThemeInterface $theme): void
+    {
+        $this->themeRoutesGenerator->generate($theme->getRoutes());
```
takeit: What about creating a "Chain Generator":
- create a compiler pass and register existing generators using custom tag (add generators to the `ThemeGeneratorClass`) 
- Create `ThemeGeneratorChain` class which will have `supports` and `generate` method where you will loop foreach generator
- add method `supports` to each generator so it supports only specific object types

It will eliminate coupling a lot and it will be easier to add new generators later by just registering a new generator with a custom tag in config, without a need to change that class and inject more dependencies to the construct.

src/SWP/Bundle/CoreBundle/Theme/TenantAwareThemeContext.php
```diff
@@ -49,11 +49,11 @@
...
      *
-     * @param TenantContextInterface   $tenantContext   Tenant context
-     * @param ThemeRepositoryInterface $themeRepository Theme repository
-     * @param CacheProvider            $cacheService    Cache Service
+     * @param TenantContextInterface             $tenantContext   Tenant context
+     * @param ReloadableThemeRepositoryInterface $themeRepository Theme repository
+     * @param CacheProvider                      $cacheService    Cache Service
      */
-    public function __construct(TenantContextInterface $tenantContext, ThemeRepositoryInterface $themeRepository, CacheProvider $cacheService)
+    public function __construct(TenantContextInterface $tenantContext, ReloadableThemeRepositoryInterface $themeRepository, CacheProvider $cacheService)
```
takeit: `reloadThemes` method is not used here so `ThemeRepositoryInterface` typing can stay here.

docs/cookbooks/templates/theme_generated_elements.rst
ahilles107: I had here before default data, but i changed it to generatedData as all of those elements are generated (articles are even using faker library). I want to have it clear that it will be created on theme installation.

docs/manual/themes/assets.rst
ahilles107: it was like that before, i just moved it to other file. and it looks ok in ToC
zrzut ekranu 2018-01-17 o 15 30 29

src/SWP/Bundle/ContentBundle/Hydrator/ArticleHydrator.php
ahilles107: i replaced this what provider was doing (https://github.com/superdesk/web-publisher/pull/424/files#diff-0f154de2d5182f063d816e8fad2a9bf4L115). But without this dependency.

src/SWP/Bundle/CoreBundle/Model/Route.php
ahilles107: We have them already in extended class

src/SWP/Bundle/CoreBundle/Theme/Processor/RequiredDataProcessor.php
ahilles107: We can think about this later if we will add new things here. For now i don't see issues with this implementation.

PR #434, opened 2018-01-30
Add support for custom fields

takeit:
| Fixed tickets             | SWP-980

TODO:

- [x] - make sure custom fields are exposed in twig
- [x] - handle media custom type fields
- [x] - docs


docs/components/TemplatesSystem/features/customTags/gimmelist.rst
```diff
@@ -71,6 +71,16 @@ To list articles from the ``Forbes`` source but without an ``AAP`` source you ca
 It will then list all articles **with** source ``Forbes`` and **without** ``AAP``.
 
 
+Listing article's custom fields:
+
+.. code-block:: twig
+
+    {% gimmelist article from articles %}
+        {{ article.title }}
+        {{ article.extra['my-custom-field'] }}
```
ahilles107: how about getter for extra fields, like: {{ article.getExtra('my-custom-field') }}?

takeit: I can add it, but I would stick to standard twig syntax because we would then need to maintain two ways of accessing these fields

ahilles107: I see that we don't have option to add getter to meta - so for now it's not possible. Let's leave it as it is.

PR #439, opened 2018-02-06
SWP-1015 - API for theme customization

takeit:
| Fixed tickets             | SWP-1015

 - [x] - API endpoint to upload theme
 - [x] - controller to stream current theme logo
 - [x] - Added theme settings
 - [x] - Reverting/restoring theme settings to defaults
 - [x] - added twig extension to render theme path
 - [x] - added twig extension to render theme settings


src/SWP/Bundle/CoreBundle/Form/Type/ThemeLogoUploadType.php
```diff
@@ -0,0 +1,67 @@
...
+     * {@inheritdoc}
+     */
+    public function buildForm(FormBuilderInterface $builder, array $options)
+    {
+        $builder
+            ->add('logo', FileType::class, [
+                'required' => true,
+                'constraints' => [
+                    new File([
+                        'maxSize' => '5120k',
```
ahilles107: why you limit it here? Imo it should be handled by php/webserwer settings

takeit: The same limit is set for theme upload zip file so I used the same.

ahilles107: 5mb, ok let's leave it like this.

PR #447, opened 2018-02-23
Generate route's slug based on route's name if not provided by default

takeit:
| Fixed tickets             | SWP-1061

TODO:
 - [x] - add migration to generate slug from name if existing routes' slug is null

### DIFF
--- a/app/migrations/Version20180223095718.php
+++ b/app/migrations/Version20180223095718.php
@@ -1,0 +1,65 @@
+<?php
+
+declare(strict_types=1);
+
+namespace SWP\Migrations;
+
+use Behat\Transliterator\Transliterator;
+use Doctrine\DBAL\Migrations\AbstractMigration;
+use Doctrine\DBAL\Schema\Schema;
+use SWP\Bundle\ContentBundle\Model\RouteInterface;
+use Symfony\Component\DependencyInjection\ContainerAwareInterface;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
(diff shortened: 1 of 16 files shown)


app/migrations/Version20180223095718.php
```diff
@@ -0,0 +1,65 @@
...
+    {
+        // this up() migration is auto-generated, please modify it to your needs
+        $this->abortIf('postgresql' !== $this->connection->getDatabasePlatform()->getName(), 'Migration can only be executed safely on \'postgresql\'.');
+
+        $entityManager = $this->container->get('doctrine.orm.default_entity_manager');
+        $routeRepository = $this->container->get('swp.repository.route');
+
+        /** @var RouteInterface $route */
+        foreach ($routeRepository->findBy(['slug' => null]) as $route) {
+            $route->setSlug(Transliterator::transliterate($route->getName()));
```
ahilles107: It will be problematic if we will modify route entity. How about doing this with clean sql?

takeit: It can be done with clean SQL but it will require some transliterate/slugify function to be executed on Postgres level explicitly. I would leave it like that for now as it does not break anything atm. Clean installation works fine too. We can discuss this further and see how it's solved in other projects in general as I saw it works fine for example in case of Sylius.

ahilles107: You can use `Transliterator` but fetch rows with sql, and call sql update query for every row.

PR #455, opened 2018-03-14
Custom routes, slug in article authors

ahilles107:
Allow to create custom route with manually provided variablePattern and requirements, add options to load authors by slug

| Fixed tickets             | ...


features/routes/management.feature
```diff
@@ -46,4 +46,32 @@ Feature: Manage Routes
...
+          "name": "Custom route",
+          "slug": "custom-route",
+          "type": "custom",
+          "variablePattern": "/{customParameter}",
+          "requirements": [
+            {
+              "key": "customParameter",
+              "value": "[a-zA-Z]+"
+            }
+		  ]
```
takeit: json format is wrong here

src/SWP/Bundle/ContentBundle/Form/Type/RouteType.php
```diff
@@ -78,7 +79,19 @@ public function buildForm(FormBuilderInterface $builder, array $options)
                 'constraints' => [
                     new GreaterThanOrEqual(0),
                 ],
-            ]);
+            ])
+            ->add('variablePattern', TextType::class, [
```
takeit: hm, it means that if I choose a different type, other than `custom` I would be still able to provide the variable pattern and requirements fields? 

We can simply listen to `type` field in the form and if on submit the `type` is `custom` we can add variable and requirements fields to the form (dynamic forms). Because for other types (`collection`, `content`) it is not needed anyway.

ahilles107: But then those field will be not visible in api doc sandbox. I will clear them if type is not custom.

ahilles107: looks like it's already done: 
```
switch ($route->getType()) {
            case RouteInterface::TYPE_CONTENT:
                $route->setVariablePattern(null);
                $route->setStaticPrefix($this->generatePath($route));
                $route->setRequirements([]);

                break;
            case RouteInterface::TYPE_COLLECTION:
                $route->setVariablePattern('/{slug}');
                $route->setStaticPrefix($this->generatePath($route));
                $route->setRequirement('slug', '[a-zA-Z0-9*\-_]+');
                $route->setDefault('slug', null);

                break;
            case RouteInterface::TYPE_CUSTOM:
                $route->setStaticPrefix($this->generatePath($route));

                break;
            default:
                throw new \InvalidArgumentException(sprintf('Route type "%s" is unsupported!', $route->getType()));
        }
```

PR #466, opened 2018-03-26
don't create 'theme' dir when installing assets, absolute dirs

takeit:

- don't create 'theme' dir when installing assets, 
- make theme assets path absolute


ahilles107:
this is not affected in 1.3-rc?

takeit:
it is. This fix should go to 1.3

PR #469, opened 2018-03-27
Missing setting in theme.json should not break pages in prod

takeit:
| Fixed tickets             | SWP-1102

### DIFF
--- a/docs/manual/themes/theme_settings.rst
+++ b/docs/manual/themes/theme_settings.rst
@@ -78,7 +78,8 @@ How to display current theme's settings in templates?
     {# app/themes/<tenant_code>/<theme_name>/views/index.html.twig #}
     {{ themeSetting('primary_font_family') }} # will print "Roboto"
 
-If the theme's setting doesn't exists an exception will be thrown with a proper message that it does not exist.
+In development environment, ff the theme's setting doesn't exists an exception will be thrown with a proper message that it does not exist.
+In production environment no exception will be thrown, the page will render normally.
 
 
 How to display current theme's settings using API?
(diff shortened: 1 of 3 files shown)


docs/manual/themes/theme_settings.rst
```diff
@@ -78,7 +78,8 @@ How to display current theme's settings in templates?
     {# app/themes/<tenant_code>/<theme_name>/views/index.html.twig #}
     {{ themeSetting('primary_font_family') }} # will print "Roboto"
 
-If the theme's setting doesn't exists an exception will be thrown with a proper message that it does not exist.
+In development environment, ff the theme's setting doesn't exists an exception will be thrown with a proper message that it does not exist.
```
ahilles107: ff ?

src/SWP/Bundle/CoreBundle/Twig/ThemeExtension.php
```diff
@@ -91,6 +99,19 @@ public function getThemeLogoPath(string $fallBackPath): string
      * @return string
      */
     public function getThemeSetting(string $setting): string
+    {
+        if ('dev' === $this->env) {
```
ahilles107: why not
```
try {
            return $this->getSetting($setting);
        } catch (\Exception $e) {
            if ('dev' !== $this->env) {
                return '';
            }
            throw $e;
        }
```

src/SWP/Bundle/CoreBundle/Twig/ThemeExtension.php
```diff
@@ -89,8 +97,23 @@ public function getThemeLogoPath(string $fallBackPath): string
...
      * @return string
+     *
+     * @throws \Exception
      */
     public function getThemeSetting(string $setting): string
+    {
+        try {
+            return $this->getSetting($setting);
+        } catch (\Exception $e) {
+            if ('dev' !== $this->env) {
```
ahilles107: maybe let's use here `'prod' ===`. As would be nice to catch those exceptions also in tests.

PR #472, opened 2018-03-28
SWP-1096 - Configure theme settings for daily and demo themes

ljubar:
added

| Fixed tickets             | SWP-1096, SWP-1005


src/SWP/Bundle/FixturesBundle/Resources/themes/DefaultTheme/theme.json
```diff
@@ -10,6 +10,45 @@
...
+                {"value": "Roboto", "label": "Roboto"}
+            ]
+        },
+        "secondary_font_family": {
+            "value": "Tahoma",
+            "type": "string",
+            "options": [
+                {"value": "Tahoma", "label": "Tahoma"},
+                {"value": "Lato", "label": "Lato"},
+                {"value": "Roboto", "label": "Roboto"}
```
aleksandarjelicic: @ljubar Would be nice to have fallback for fonts like sans-serif, just in case of google font error.

src/SWP/Bundle/FixturesBundle/Resources/themes/DefaultTheme/views/base.html.twig
```diff
@@ -34,14 +34,38 @@
   {% block amphtml %}{% endblock %}
   {% block stylesheets %}
   <!-- CSS: implied media="all" -->
+  <link href="https://fonts.googleapis.com/css?family=Oswald&subset=latin,latin-ext" rel="stylesheet">
```
ahilles107: i think that we should load here font set in settings

PR #635, opened 2018-09-25
SF4 part 1 - env vars

takeit:
| Fixed tickets             | SWP-1328

- introduced env vars, removed `parameters.yml`
- use `%kernel.project_dir%` instead of deperecated `%kernel.root_dir%` parameter
- moved packages config to `config` dir
- moved front controller to `public/index.php`, removed `app.php, app_dev.php, app_test.php` front controllers
- moved `app/Resources` templates to `templates` dir
- moved `AppKernel.php` to `src/Kernel.php`
- created `var` dir for logs and cache


app/console
```diff
@@ -3,28 +3,36 @@
...
 use Symfony\Bundle\FrameworkBundle\Console\Application;
 use Symfony\Component\Console\Input\ArgvInput;
-use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Debug\Debug;
-
-// if you don't want to setup permissions the proper way, just uncomment the following PHP line
-// read http://symfony.com/doc/current/setup.html#checking-symfony-application-configuration-and-setup
-// for more information
-//umask(0000);
+use Symfony\Component\Dotenv\Dotenv;
```
ahilles107: it will break prod, as dotenv is in require-dev section

config/packages/fos_http_cache.yaml
```diff
@@ -9,7 +12,7 @@ fos_http_cache:
     proxy_client:
         varnish:
             http:
-                servers: "%cache_servers%"
+                servers: ["localhost"]
```
ahilles107: why like that? Personally i have varnish on other host than localhost.

config/packages/prod/doctrine.yml
```diff
@@ -1,12 +1,8 @@
...
     dbal:
-        driver:   pdo_pgsql
-        host:     '%env(DATABASE_HOST)%'
-        port:     '%env(DATABASE_PORT)%'
-        dbname:   '%env(DATABASE_NAME)%'
-        user:     '%env(DATABASE_USER)%'
-        password: '%env(DATABASE_PASSWORD)%'
+        driver: pdo_pgsql
+        url: '%env(resolve:DATABASE_URL)%'
+        server_version: '%env(resolve:DATABASE_SERVER_VERSION)%'
```
ahilles107: in new sf4 projects i pass server version as part of database url: `?charset=utf8&serverVersion=10`

public/index.php
```diff
@@ -0,0 +1,38 @@
+<?php
```
ahilles107: This will be big bc break for users, and our infrastructure team. Need to be described in UPGRADE.md

config/packages/fos_http_cache.yaml
takeit: it's still WIP, hardcoded it to make it work for now

app/console
takeit: it won't because the Dotenv class won't be constructed in prod at all if the `APP_ENV` var is set, and it must be set to `prod` in prod env

.travis.yml
```diff
@@ -30,7 +30,6 @@ env:
 matrix:
   allow_failures:
   - php: nightly
-  - php: 7.2
```
ahilles107: 👍

PR #660, opened 2018-11-08
Fix/webhooks multiple articles update fix

ahilles107:
| Fixed tickets             | 


src/SWP/Bundle/CoreBundle/EventSubscriber/WebhookEventsSubscriber.php
```diff
@@ -66,33 +68,32 @@ public static function getSubscribedEvents(): array
         return $subscribedEvents;
     }
 
-    public function handleEvent(Event $event): void
+    public function handleEvent(Event $event, string $dispatcherEventName, EventDispatcherInterface $dispatcher): void
```
takeit: why do you pass dispatcher service here instead of injecting it as a dependency inside `WebhookEventsSubscriber` class?

ahilles107: It's passed there by default by Dispatcher already.

PR #701, opened 2019-01-17
Improvement/speed up content publishing

ahilles107:
| Fixed tickets             | 


src/SWP/Bundle/CoreBundle/Context/ScopeContext.php
```diff
@@ -14,19 +14,66 @@
...
+
+        return parent::getScopeOwner($scope);
+    }
+
+    private function lazyLoad(string $scope)
+    {
+        if (in_array($scope, self::NEW_SCOPES)) {
+            $tenant = $this->tenantContent->getTenant();
+            if ($tenant instanceof SettingsOwnerInterface) {
+                if (ScopeContextInterface::SCOPE_TENANT === $scope) {
```
takeit: can this be somehow simplified? It's really hard to read that, what about using a switch statement instead?

PR #724, opened 2019-02-08
[feature] add option to plan adding article to selected conent list when it's published

ahilles107:
| Tests pass?               | yes|
| Fixed tickets             | 


src/SWP/Bundle/CoreBundle/Form/Type/ContentListPositionType.php
```diff
@@ -0,0 +1,50 @@
...
+use Symfony\Component\OptionsResolver\OptionsResolver;
+
+final class ContentListPositionType extends AbstractType
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function buildForm(FormBuilderInterface $builder, array $options)
+    {
+        $builder->add('id', IntegerType::class);
```
takeit: IMO you should write and use here a custom type, e.g. `ContentListSelectorType` which will use data transformer to transform content list id to Content List object and vice versa. Because at the moment you can type whatever id you want here and no validation will be triggered.

ahilles107: But for what when in database we keep it as serialized array? We don't need at this moment those content lists objects at all.

takeit: for validation purposes, to not enter not existing content list id and do not store it in the database. Why would you want to store not existing content list id?

takeit: or just add a custom validator to check if the list with provided id exists if these content lists objects are not needed

PR #733, opened 2019-02-19
implement simple events support in meta class, implement dynamic loading of article content lists

ahilles107:
| Fixed tickets             | SWP-1471

### DIFF
--- a/app/Resources/meta/article.yml
+++ b/app/Resources/meta/article.yml
@@ -59,4 +59,7 @@ properties:
     commentsCount:
         description: "Number of comments (managed by external system)"
         type: integer
+    contentLists:
+        description: "An array with content lists to witch this article is assigned"
+        type: array
 to_string: title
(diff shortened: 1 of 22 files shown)


app/Resources/meta/article.yml
```diff
@@ -59,4 +59,7 @@ properties:
     commentsCount:
         description: "Number of comments (managed by external system)"
         type: integer
+    contentLists:
+        description: "An array with content lists to witch this article is assigned"
```
takeit: `which`

src/SWP/Component/TemplatesSystem/Gimme/Event/MetaEvent.php
```diff
@@ -0,0 +1,60 @@
...
+
+    private $propertyName;
+
+    public function __construct($data, string $propertyName)
+    {
+        $this->data = $data;
+        $this->propertyName = $propertyName;
+    }
+
+    public function getData()
```
takeit: return types are missing for some of the functions in this class

ahilles107: data can be mixed (object, array) so no option to use return type for it.

PR #740, opened 2019-02-27
Related articles

takeit:
| Fixed tickets             | SWP-1338

- added related article loader to load related articles in twig
- API endpoint to load related articles by article id

TODO:
- [x] docs (RelatedArticleLoader)
- [x] test article correction with related articles (remove, add related items)
- [x] API endpoint to check where related articles were published


src/SWP/Bundle/ContentBundle/Controller/RelatedArticleController.php
```diff
@@ -0,0 +1,76 @@
...
+     *     description="List related articles",
+     *     statusCodes={
+     *         200="Returned on success.",
+     *         500="Unexpected error."
+     *     },
+     *     filters={
+     *         {"name"="sorting", "dataType"="string", "pattern"="[updatedAt]=asc|desc"}
+     *     }
+     * )
+     * @Route("/api/{version}/content/related/articles/{id}", options={"expose"=true}, defaults={"version"="v1"}, name="swp_api_content_list_related_articles", requirements={"id"="\d"})
```
ahilles107: Why not `/api/{version}/content/articles/{id}/related` ? Imo will be more consistent with with current articles endpoint.

src/SWP/Bundle/ContentBundle/Doctrine/ORM/RelatedArticleRepository.php
```diff
@@ -0,0 +1,77 @@
...
+use SWP\Component\Common\Pagination\PaginationData;
+
+class RelatedArticleRepository extends EntityRepository implements RelatedArticleRepositoryInterface
+{
+    public function getByCriteria(Criteria $criteria, array $sorting): array
+    {
+        $queryBuilder = $this->createQueryBuilder('ra');
+
+        $this->applyCustomCriteria($queryBuilder, $criteria);
+        $this->applyCriteria($queryBuilder, $criteria, 'ra');
```
ahilles107: it can be replaced with 
`getQueryByCriteria` from trait.

src/SWP/Bundle/ContentBundle/Model/Article.php
```diff
@@ -126,6 +126,9 @@ class Article implements ArticleInterface
      */
     protected $slideshows;
 
+    /**
+     * Article constructor.
```
ahilles107: looks like wrong docs

src/SWP/Bundle/ContentBundle/Controller/RelatedArticleController.php
takeit: because there is already a route `/api/{version}/content/articles/{id}` where id param has `.+` requirement so route `api/v1/content/articles/7/related` is matched by `/api/{version}/content/articles/{id}` and not `/api/{version}/content/articles/{id}/related`. We should either limit id in `api/{version}/content/articles/{id}` route to `\d` or manually configure route in yaml and set `/api/{version}/content/articles/{id}/related` route to be mached before `api/{version}/content/articles/{id}`. But since we use annotations we should stick with that, otherwise we move alll routing definition to yaml files IMO.

src/SWP/Bundle/ContentBundle/Doctrine/ORM/RelatedArticleRepository.php
takeit: it can't be because I won't have access to QueryBuilder to which I want to apply custom criteria and I can't add more criteria to query builder after limiting is set.

src/SWP/Bundle/ContentBundle/Controller/RelatedArticleController.php
ahilles107: In case of annotation order of actions in controller set's their priority - this same like in yaml. If action will be before normall article get then it will be checked first. That should be enough.

src/SWP/Bundle/ContentBundle/Doctrine/ORM/RelatedArticleRepository.php
ahilles107: ahh... it returns query not query builder.

src/SWP/Bundle/ContentBundle/Controller/RelatedArticleController.php
takeit: but this is a separate controller and routes priorities are set based on controller order by name

takeit: I can move this action to ArticleController but it makes sense to keep it in a separate one to not mix stuff IMO

src/SWP/Bundle/ContentBundle/Model/Article.php
```diff
@@ -126,8 +126,14 @@ class Article implements ArticleInterface
...
     protected $slideshows;
 
+    /**
+     * @var bool
+     */
     private $isTimestampableCanceled = false;
 
+    /**
+     * Article constructor.
+     */
```
ahilles107: this can be removed ;)

src/SWP/Bundle/ContentBundle/Model/RelatedArticle.php
```diff
@@ -0,0 +1,67 @@
...
+namespace SWP\Bundle\ContentBundle\Model;
+
+use SWP\Component\Common\Model\SoftDeletableTrait;
+use SWP\Component\Common\Model\TimestampableTrait;
+
+class RelatedArticle implements RelatedArticleInterface
+{
+    use TimestampableTrait, SoftDeletableTrait;
+
+    protected $id;
```
ahilles107: we have persistable interface and trait

src/SWP/Bundle/ContentBundle/Model/Article.php
takeit: damn I thought I did it :P

PR #759, opened 2019-03-14
slideshows over webhooks improvements

takeit:
SWP-1558


src/SWP/Bundle/ContentBundle/KeyGenerator/MetaKeyGenerator.php
```diff
@@ -32,8 +32,11 @@ public function generateKey($meta): string
             $keyElements = [];
 
             if ($value instanceof TimestampableInterface) {
-                $date = null !== $value->getUpdatedAt() ? $value->getUpdatedAt() : $value->getCreatedAt();
-                $keyElements[] = $date->getTimestamp();
+                $date = $value->getUpdatedAt() ?? $value->getCreatedAt();
+
+                if (null !== $date) {
```
ahilles107: But created at should be allways set. So with 
```$date = null !== $value->getUpdatedAt() ? $value->getUpdatedAt() : $value->getCreatedAt();``` `$date` will never be `null`.

File: src/SWP/Bundle/CoreBundle/Serializer/CachedJMSSerializer.php [hunk 2]
```diff
@@ -59,7 +59,7 @@ public function deserialize($data, $type, $format)
     private function getCacheKey($data, string $type = null, string $format = null): ?string
     {
         if ($data instanceof TimestampableInterface && $data instanceof PersistableInterface) {
-            $keyElements = [$data->getId(), $data->getCreatedAt()->getTimestamp()];
+            $keyElements = [$data->getId(), $data->getCreatedAt() ? $data->getCreatedAt()->getTimestamp() : null];
```
ahilles107: This same here, we should fix issue with created at set to null not this one.

PR #771, opened 2019-04-03
return array of objects instead of object of objects

takeit:
| Fixed tickets             | SWP-1581


takeit:
fixed in https://github.com/superdesk/web-publisher/pull/777

PR #835, opened 2019-05-15
[Feature] Generate direct media urls

ahilles107:
- [x] Bug fix
  - [x] New feature
  - [ ] BC breaks
  - [ ] Deprecations
  - [x] Changelog update (if required)

Fixed tickets : 
License: AGPLv3


src/SWP/Bundle/ContentBundle/Resolver/AssetLocationResolver.php
```diff
@@ -0,0 +1,62 @@
...
+    {
+        $this->mainAdapter = $mainAdapter;
+        $this->awsBucket = $awsBucket;
+        $this->localDirectory = $localDirectory;
+        $this->awsClient = $awsClient;
+    }
+
+    public function getAssetUrl(FileInterface $file): string
+    {
+        if ('aws_adapter' === $this->mainAdapter) {
```
takeit: you can put that into a separate class `AssetUrlGenerator` with a separate interface and one method `getUrl` and inject that service here.

Then it should be possible to register custom generator, for example, AwsUrlGenerator, LocalUrlGenerator (by adding tags `swp.asset_url_generator` to it) and in the future, if you would want to use another storage you could easily add a new class to support it without making changes to the existing code. Because right now it's all hardcoded here.

PR #842, opened 2019-05-22
[feature] WEBP images versions

ahilles107:
Create webp versions of article images, add support for them in rendition meta, handle url generation

  - [ ] Bug fix
  - [x] New feature
  - [ ] BC breaks
  - [ ] Deprecations
  - [x] Changelog update (if required)

Fixed tickets : SWP-1605
License: AGPLv3


app/Resources/meta/imageRendition.yml
```diff
@@ -11,4 +11,7 @@ properties:
     height:
         description: "Rendition height"
         type: int
+    convertedToWebp:
```
takeit: if it's a boolean value, then the `isConvertedToWebp` sounds much better

src/SWP/Bundle/ContentBundle/EventListener/ProcessArticleMediaListener.php
```diff
@@ -57,7 +57,7 @@ function ($entry) use ($guids) {
 
             if (null !== ($packageItems = $packageItem->getItems()) && 0 !== $packageItems->count()) {
                 $packageItems = $packageItem->getItems()->filter(
-                    function ($entry) use ($guids) {
+                    static function ($entry) use ($guids) {
```
takeit: 👍

src/SWP/Bundle/ContentBundle/Model/ImageRendition.php
```diff
@@ -118,4 +93,9 @@ public function setName(string $name): void
     {
         $this->name = $name;
     }
+
+    public function isConvertedToWebp(): bool
+    {
+        return $this->getImage()->haveVariant('webp');
```
takeit: it should be `hasVariant`. You can assign `webp` to a const in `ImageInterface` as it's used in more places.

app/Resources/meta/imageRendition.yml
ahilles107: but then getter used by Meta class would be `isIsConvertedToWebp`. I will need to add some more logic to meta class getters generator.

PR #869, opened 2019-06-16
add etag to response

mrhein:


ahilles107:
Hmm... How about moving it to event listener? Adding it to front controller isn't best idea. In this case it will not work in dev env. In 2.0 we have just one front so it solves that issue but still - doesn't look like correct approach - instead that we should modify response on `kernel.response` event.

ahilles107:
@mrhein - how it looks? it's still priority for You?

ahilles107:
@mrhein ping ;)

PR #914, opened 2019-07-24
Read content type from route

mehdignu:
- [ ] Bug fix
  - [x] New feature
  - [ ] BC breaks
  - [ ] Deprecations
  - [ ] Changelog update (if required)

Fixed tickets : comma separated list of tickets fixed by the PR
License: AGPLv3

### DIFF
--- a/src/SWP/Bundle/ContentBundle/EventListener/MimeTypeListener.php
+++ b/src/SWP/Bundle/ContentBundle/EventListener/MimeTypeListener.php
@@ -49,7 +49,7 @@ class MimeTypeListener
                         $extension = $event->getRequest()->get($matches[1]);
                     }
                 }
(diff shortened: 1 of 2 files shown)


ahilles107:
@mehdignu plmease apply styleci proposed changes and we can merge it (https://github.styleci.io/analyses/zGWE4E?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) 

THANKS a lot for contribution!

ahilles107:
@mehdignu actually there is one more thing. We do have tests for this class `MimeTypeListenerTest.php`.  Please add test case for it (create route like expected one, and set parameters in requests) to verify that it works like it should. Then we will be covered in futer changes here.

ahilles107:
@mehdignu thanks a lot for Your work! All tests passed, and code looks good! We have now content type set also on custom routes with variable patterns ;)

PR #939, opened 2019-08-22
Feature media is unset when an article is updated or corrected

takeit:
- [x] Bug fix
  - [ ] New feature
  - [ ] BC breaks
  - [ ] Deprecations
  - [ ] Changelog update (if required)

Fixed tickets: SWP-1726
License: AGPLv3

Issue:
When the existing article with feature media has been corrected or updated, the feature media has been deleted from the article.


takeit:
FYI @mehdignu

PR #944, opened 2019-08-30
handle package correction for update without original package in publisher

ahilles107:
License: AGPLv3

### DIFF
--- a/src/SWP/Bundle/CoreBundle/Consumer/ContentPushConsumer.php
+++ b/src/SWP/Bundle/CoreBundle/Consumer/ContentPushConsumer.php
@@ -178,11 +178,10 @@ class ContentPushConsumer implements ConsumerInterface
 
     protected function findExistingPackage(PackageInterface $package)
     {
-        $existingPackage = null;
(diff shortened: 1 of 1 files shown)


src/SWP/Bundle/CoreBundle/Consumer/ContentPushConsumer.php
```diff
@@ -178,15 +178,15 @@ public function doExecute(TenantInterface $tenant, PackageInterface $package): i
 
     protected function findExistingPackage(PackageInterface $package)
     {
-        $existingPackage = null;
         if (null === $package->getEvolvedFrom()) {
-            $existingPackage = $this->packageRepository->findOneBy([
-                'evolvedFrom' => $package->getGuid(),
-            ]);
+            $existingPackage = $this->packageRepository->findOneBy(['guid' => $package->getGuid()]);
+        } else {
```
takeit: you can use null coalescing operator here `['guid' => $package->getEvolvedFrom() ?? $package->getGuid()]`

PR #951, opened 2019-09-10
remove not needed seo metadata listener

ahilles107:
## Proposed Changes

  - remove listener for seo metadata - it wasn't doing anything anyhow

License: AGPLv3


src/SWP/Bundle/CoreBundle/EventListener/AssignSeoMetadataListener.php
```diff
@@ -1,50 +0,0 @@
...
-            return;
-        }
-
-        $article->setSeoMetadata(null);
-
-        $newSeoMetadata = clone $seoMetadata;
-        $newSeoMetadata->setPackageGuid(null);
-        $newSeoMetadata->setId(null);
-
-        $article->setSeoMetadata($seoMetadata);
```
ahilles107: We were setting back originally fetched seo metadata - not the cloned one. And setting cloned one creates another issues with media id's (as those are not cloned).

ahilles107:
actually it's wrong because without this listener seo metadata is not attached to article at all.

ahilles107:
many articles created for this same package guid and this:
```   oneToOne:
        seoMetadata:
            targetEntity: SWP\Bundle\ContentBundle\Model\ArticleSeoMetadataInterface
            joinColumn:
                name: seo_metadata_id
                referencedColumnName: id
            cascade: ['all']```

metadata is related to package not article so in case of articles in two tenants it will fail.

PR #952, opened 2019-09-11
fetch/serialize only needed properties for packages endpoint

ahilles107:
Fixes - long time of loading packages endpoint 

## Proposed Changes

  - Don't load and serialize not needed relations/properties on `api/v2/packages/` endpoint. 
 
License: AGPLv3

### DIFF
--- a/features/content_lists/add_article_to_manual_content_list_when_published.feature
+++ b/features/content_lists/add_article_to_manual_content_list_when_published.feature
@@ -5,7 +5,7 @@ Feature: Add article to manual content list when published
   As a HTTP Client
   I want to be able to publish article and see it in content lists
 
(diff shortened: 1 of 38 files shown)


features/content_lists/add_article_to_manual_content_list_when_published.feature
```diff
@@ -0,0 +1,81 @@
+@disable-fixtures
+@content_lists
+Feature: Add article to manual content list when published
+  In order to add article to content list when it's published
+  As a HTTP Client
+  I want to be able to publish article and see it in content lists
+
+  Scenario: Push new article and add it to automatic content lists
```
takeit: it should be manual, not automatic because later you create a manual content list

PR #956, opened 2019-09-24
fire post published event in article publisher

ahilles107:
## Proposed Changes

  - Fire POST_PUBLISHED after article publication

License: AGPLv3


features/packages/publish_package_with_tenant_rules.feature
```diff
@@ -115,7 +115,9 @@ Feature: Publish article based on configured organization and tenant rules
     Then the response status code should be 200
     And the JSON node "status" should be equal to "published"
 
+
     Then I am authenticated as "test.user"
+    And I wait 1 second
```
ahilles107: i wait here because we gat data from elastic and sometimes it takes time

PR #958, opened 2019-09-25
log this same (and more) data to filesystem and graylog

ahilles107:
Fixes:

Issue with production failing when correct graylog instance data are not set

## Proposed Changes

  - log this same data to graylog and filesystem 
  - log app channel data only when error level message i raised

License: AGPLv3

### DIFF
--- a/config/packages/prod/monolog.yaml
+++ b/config/packages/prod/monolog.yaml
@@ -20,23 +20,15 @@ monolog:
         - swp_asset_download
         - swp_content_push
         - swp_image_conversion
+        - swp_rule
+        - swp_validators
     handlers:
-        main:
-            type:         fingers_crossed
-            action_level: error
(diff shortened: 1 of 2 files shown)


config/packages/prod/monolog.yaml
```diff
@@ -20,27 +20,36 @@ monolog:
...
-            level: error
-            channels: ["!swp_send_webhook", "!swp_content_push", "!swp_image_conversion", "!swp_asset_download", "!swp_rule", "!swp_validators"]
         console:
             type:  console
-        gelf:
-            level: debug
-            type: service
-            id: monolog.gelf_handler
+        custom_channels:
+            type: fingers_crossed
```
ahilles107: this may be changed to `buffer` maybe but `fingers_crossed` with this `action_level` also works.

PR #965, opened 2019-09-26
allow to sort articles by page views in elastic search

takeit:
This will require to repopulate the ES index

Fixes SWP-1358

License: AGPLv3


takeit:
@tomaszrondio please take a look

PR #986, opened 2019-11-06
Fdl 548 - modified date when updating a published article

mehdignu:
Fixes #

## Proposed Changes
the article's modified date should be changed according to the type of change:

- if grammar correction the modified date should not be updated.
- if republishing modified date should be updated.
  
License: AGPLv3


ahilles107:
I see two changes here, can You explain both of them?

ahilles107:
And we don't use `modifiedAt` in templates. We use `updateAt` and it's automatically updated. What we need to do here is to disable that functionality in some cases - not introduce new property controlled manually. We have something like that  here for timestampable: `SWP\Bundle\ContentBundle\EventListener\TimestampableListener`

mehdignu:
@ahilles107 sounds good i will work on that. the other change "length" is for the images size

ahilles107:
@mehdignu image length should be in seperate pull request so we could merge it without waiting on changes here.

mehdignu:
alright.

PR #988, opened 2019-11-07
[FEATURE] modified date when updating a published article

mehdignu:
Fixes #

## Proposed Changes

there are two versions for updating a published article.
1) only some grammar corrections --> no republishing/ every timestamp stays the same (modified date will not be updated) no changes for date!
2) content adjustments or changes --> republishing / time stamp for modified date changes because of the new content inside the article --> also in FE there should be the modified date displayed

License: AGPLv3

### DIFF
--- a/src/SWP/Bundle/ContentBundle/EventListener/ArticlePublishListener.php
+++ b/src/SWP/Bundle/ContentBundle/EventListener/ArticlePublishListener.php
@@ -44,7 +44,7 @@ final class ArticlePublishListener
     {
         $article = $event->getArticle();
 
-        if (isset($article->getExtra()['update_date'])) {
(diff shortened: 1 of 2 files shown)


src/SWP/Bundle/ContentBundle/EventListener/ArticlePublishListener.php
```diff
@@ -44,6 +44,10 @@ public function publish(ArticleEvent $event)
     {
         $article = $event->getArticle();
 
+        if (isset($article->getExtra()['update_date'])) {
```
ahilles107: Let's make it more verbose. Change `update_date` to `dont_update_date`.

ahilles107:
Adjust phpspec tests and add new behat feature to cover that case with test.

PR #989, opened 2019-11-07
[FEATURE] - Add images length to the Image model

mehdignu:
Fixes #

## Proposed Changes

  - add images length to the Images model

License: AGPLv3


src/SWP/Bundle/ContentBundle/Factory/ORM/MediaFactory.php
```diff
@@ -153,6 +153,12 @@ private function downloadAsset(string $url, string $media, string $mimetype): Fi
             [$width, $height] = \getimagesize($uploadedFile->getRealPath());
             $file->setWidth($width);
             $file->setHeight($height);
+            $size = \filesize($uploadedFile->getRealPath());
+            $size = $size / 1024;
+            $size = (string) number_format($size);
+            if (null !== $size) {
```
ahilles107: There is a bug here. On line 158 you cast value to string. `null` casted to string will be `''`. So this if will never catch issue with number_format return value. That's one thing. 

Second - You want to keep this value as string - can't we keep it with `int` or `float` ?

ahilles107:
We need tests for it before merging. This can be added probably to some exisiting functionall php unit test.

mehdignu:
@ahilles107 ping

PR #993, opened 2019-11-12
expose article place

ahilles107:
Fixes #SWP-1667

## Proposed Changes

  - add option to get first places deifined in article metadata with simple short syntax article.place.city

License: AGPLv3

### DIFF
--- a/app/Resources/meta/article.yml
+++ b/app/Resources/meta/article.yml
@@ -65,4 +65,7 @@ properties:
     seoMetadata:
         description: "SEO metadata"
         type: Meta
+    place:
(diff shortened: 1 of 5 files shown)


app/Resources/meta/article.yml
```diff
@@ -65,4 +65,7 @@ properties:
     seoMetadata:
         description: "SEO metadata"
         type: Meta
+    place:
+        description: "First place array from place metadata"
```
takeit: I don't get this description. I think it should be something like: The first place from the article's metadata

PR #994, opened 2019-11-13
change author biography from string to text in database

ahilles107:
Fixes # SWP-1667

## Proposed Changes

  - add missing migration for removing article_events table
  - add new image field - length

License: AGPLv3

### DIFF
--- a/src/SWP/Bundle/CoreBundle/Migrations/2019/11/Version20191113081727.php
+++ b/src/SWP/Bundle/CoreBundle/Migrations/2019/11/Version20191113081727.php
@@ -1,0 +1,102 @@
+<?php
+
+declare(strict_types=1);
+
+namespace SWP\Migrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
+/**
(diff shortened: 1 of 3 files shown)


src/SWP/Bundle/CoreBundle/Migrations/2019/11/Version20191113081727.php
```diff
@@ -0,0 +1,102 @@
...
+
+namespace SWP\Migrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
+/**
+ * Auto-generated Migration: Please modify to your needs!
+ */
+final class Version20191113081727 extends AbstractMigration
```
takeit: the migration has some extra not needed sql queries

ahilles107: which ones?

takeit: all of them besides the ones related to `biography` field

PR #996, opened 2019-11-14
[SWP-1814] handle new superdesk ways for unpublishing content

ahilles107:
Fixes #SWP-1814

## Proposed Changes

  - remove canceled package listener - move that functionallity to updatedPackageListener

License: AGPLv3


features/02_articles/unpublish_from_superdesk.feature
```diff
@@ -0,0 +1,129 @@
...
+            {
+              "tenant":"123abc",
+              "published":true,
+              "route": 1
+            }
+          ]
+      }
+     """
+
+#    When I go to "/test/lorem"
```
takeit: this can stay probably to check if the content is published before unpublishing

PR #1008, opened 2019-11-28
Add env variable to customize S3 endpoint

kottkrig:
# What does this PR do?

It adds a env variable to customize the S3 endpoint so that publisher can use S3-like services.

# Why was this PR needed?

Previously, the endpoint defaulted to AWS S3 which made it impossible to use with other S3-compliant services.


kottkrig:
Is there a way to disable an env variable if it's empty or not set? It seems that the tests doesn't approve of an empty endpoint.

ahilles107:
i will check this broken tests - it looks like not connected to this change.

src/SWP/Bundle/ContentBundle/Resources/config/app/config.yml
```diff
@@ -6,6 +6,7 @@ parameters:
     env(FS_AWS_S3_REGION): ''
     env(FS_AWS_S3_KEY): ''
     env(FS_AWS_S3_SECRET): ''
+    env(FS_AWS_S3_ENDPOINT): null
```
ahilles107: @kottkrig please set here default aws s3 endpoint. 

```Endpoints must be full URIs and include a scheme and host ```

Comment (takeit) [reply src/SWP/Bundle/ContentBundle/Resources/config/app/config.yml]: It can be just an empty string according to this: https://github.com/guzzle/psr7/blob/master/src/Uri.php#L77

@kottkrig I saw you set it to `env(FS_AWS_S3_ENDPOINT)= ''` before then changed to `env(FS_AWS_S3_ENDPOINT): null`, the empty string didn't work?

Comment (kottkrig) [reply src/SWP/Bundle/ContentBundle/Resources/config/app/config.yml]: I believe that I got the same error with an empty string but I can try it again.

Comment (kottkrig) [reply src/SWP/Bundle/ContentBundle/Resources/config/app/config.yml]: Yes. The same error with an empty string:

> Endpoints must be full URIs and include a scheme and host

Comment (takeit) [reply src/SWP/Bundle/ContentBundle/Resources/config/app/config.yml]: yep, to solve this problem, you can wrap the instantiation of `Aws\S3\S3Client` in a new class and pass the arguments (`version`, `region` etc.) into it, so if the `env(FS_AWS_S3_ENDPOINT)= ''` env var is an empty string then `endpoint` argument should not be set in the `Aws\S3\S3Client` at all.
You can try this https://symfony.com/doc/4.3/service_container/factories.html

Comment (kottkrig) [reply src/SWP/Bundle/ContentBundle/Resources/config/app/config.yml]: Thank you. I will have a go at it in a couple of days.

Comment (ahilles107) [reply src/SWP/Bundle/ContentBundle/Resources/config/app/config.yml]: Guys, why not setting default aws endpoint as default value?

Comment (takeit) [reply src/SWP/Bundle/ContentBundle/Resources/config/app/config.yml]: it's because different regions can have different endpoints https://docs.aws.amazon.com/general/latest/gr/s3.html and the default value of `endpoint` must be valid URL or not set at all in order to fall back to a default endpoint in the PHP client. It's should be optional to set the value of `endpoint`

Comment (ahilles107) [reply src/SWP/Bundle/ContentBundle/Resources/config/app/config.yml]: @takeit - then yes - we need some special logic in wrapper set before the client.

**Comment by ahilles107** [comment]
Hey @kottkrig - did You had time to look at it?

**Comment by kottkrig** [comment]
Not yet unfortunately. But I make sure to have a look at it tomorrow.

File: src/SWP/Bundle/ContentBundle/Factory/S3ClientFactory.php [hunk 2]
```diff
@@ -0,0 +1,37 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of the Superdesk Web Publisher Content Bundle.
+ *
+ * Copyright 2016 Sourcefabric z.ú. and contributors.
```
ahilles107: please change it to 2019

src/SWP/Bundle/ContentBundle/Factory/S3ClientFactory.php
```diff
@@ -0,0 +1,37 @@
...
+ *
+ * For the full copyright and license information, please see the
+ * AUTHORS and LICENSE files distributed with this source code.
+ *
+ * @copyright 2016 Sourcefabric z.ú
```
ahilles107: also here ;)

src/SWP/Bundle/ContentBundle/Resources/config/services.yml
```diff
@@ -326,22 +326,24 @@ services:
         arguments:
             - '%env(json:BLACKLISTED_KEYWORDS)%'
 
-    Aws\S3\S3Client:
+    S3Client:
```
ahilles107: we are trying to use full FQCN for service names. Please change it from `S3Client` to `SWP\Bundle\ContentBundle\Factory\S3ClientFactory`.

ahilles107:
💯 ;)

kottkrig:
Thanks for the feedback. I'll have a look at these things first thing tomorrow.

ahilles107:
@kottkrig thanks a lot for this work! I will fix this issue on latest deps on travis in second PR (we need to stick for now to older phrophecy release).

kottkrig:
@ahilles107 do you have an ETA on when you will be able to merge this onto the 2.0-branch?

(Or, even better, do you have an ETA on when it might be released officially?)

ahilles107:
@kottkrig  - it will not be merged to 2.0.x at all. From master branch we will create soon 2.1.0. First beta should be ready next week, with stable release planned on March/April.

kottkrig:
@ahilles107 thanks for the info. 👍

PR #1009, opened 2019-12-02
Restrict access to articles based on geo-location

takeit:
## Proposed Changes

This PR introduces a new feature which allows restricting access to articles based on geo-location and article's place metadata.

It uses MaxMind GeoIP2 database where we lookup client IP address and find the proper country or city. The database has to be downloaded first using the command: `php bin/console swp:geoip:db:update`.

When this feature is enabled, the reader's IP address is read on the visited article page.
Then, the IP address is checked in the GeoIP2 database to check which country or state it comes from.
The GeoIP database has to be downloaded first.
If the article place metadata matches the IP country or state, access to the articles is denied.

The GeoIP features are disabled by default.
To enable the GeoIP features you have to set the ``GEO_IP_ENABLED`` environment variable to ``true`` (``GEO_IP_ENABLED=true``)
in your .env.local file.

This PR adds also a new component which is called `Archiver` with basic unarchive methods for different formats. At the moment only zip and gzip unarchiving is supported.

TODO:
  - [x] - create a new GeoIpReader in `CoreBundle` and cache calls to the GeoIP2 database
  - [x] - implement the command to download the MaxMind GeoIP2 database
  - [x] - create a new interface to which will be used in the GeoIpReader instead of using `ArticleInterface` and passing the whole article's object. The needed fields are only `country` and `city`.

License: AGPLv3


src/SWP/Bundle/CoreBundle/Resources/config/services.yml
```diff
@@ -766,3 +766,6 @@ services:
 
     SWP\Bundle\CoreBundle\Hydrator\PackageHydratorInterface:
         alias: SWP\Bundle\CoreBundle\Hydrator\PackageHydrator
+
+    SWP\Bundle\CoreBundle\GeoIp\CachedGeoIpChecker:
+        autowire: true
```
ahilles107: missing new line at end of the file

src/SWP/Bundle/GeoIPBundle/DependencyInjection/Configuration.php
```diff
@@ -0,0 +1,46 @@
...
+     * {@inheritdoc}
+     */
+    public function getConfigTreeBuilder()
+    {
+        $treeBuilder = new TreeBuilder('swp_geo_ip');
+        $treeBuilder->getRootNode()
+            ->children()
+                ->scalarNode('database_url')
+                    ->cannotBeEmpty()
+                    ->defaultValue('https://geolite.maxmind.com/download/geoip/database/GeoLite2-City.mmdb.gz')
```
ahilles107: can we keep it as constraint somewhere? It's second time when i see this url.

src/SWP/Component/Archiver/Archiver/ZipArchiver.php
```diff
@@ -0,0 +1,44 @@
+<?php
```
ahilles107: We should use it also in `OrganizationAwareThemeUploader`. It requires a little more functionality, but at the end does this same.

src/SWP/Component/GeoIP/Checker/GeoIPChecker.php
```diff
@@ -0,0 +1,53 @@
...
+{
+    /** @var ReaderInterface */
+    private $reader;
+
+    public function __construct(ReaderInterface $reader)
+    {
+        $this->reader = $reader;
+    }
+
+    public function isGranted(string $ipAddress, Place $place): bool
```
ahilles107: i don't get it. `isGranted` returns `false` when state, or country matches. And in other cases returns `true`?

src/SWP/Component/Archiver/Archiver/ZipArchiver.php
takeit: another PR

src/SWP/Component/GeoIP/Checker/GeoIPChecker.php
takeit: yes, because if the article country equals to the country based on IP it should not allow seeing the article.

src/SWP/Bundle/GeoIPBundle/DependencyInjection/Configuration.php
takeit: it's only here, maybe you saw it in the docs?

src/SWP/Component/GeoIP/Checker/GeoIPChecker.php
ahilles107: But in places we should provide allowed countries/states - not excluded.

src/SWP/Bundle/GeoIPBundle/DependencyInjection/Configuration.php
ahilles107: yep

src/SWP/Component/GeoIP/Checker/GeoIPChecker.php
takeit: we agreed that when the place in SD will be defined, readers from these "places" will not be able to read the article.

PR #1017, opened 2019-12-12
Redirect routes

takeit:
## Proposed Changes

This PR adds a new feature called Redirect Routes. It uses the existing Routes implementation.

1. It's possible to create a redirect route which will redirect existing Publisher's route to another existing route, e.g. `/politics` -> `/news`.
2. It's possible to define the redirect type, either it's permanent (301 status code) or not (302 status code).
3. It's possible to type custom redirect expression and redirect to an external URL, e.g. set redirect route name to `/random-string` and define URI to `https://google.com`, then when visiting `http://example.com/random-string` URL, the visitor will be redirected to `https://google.com` with a proper status code.

License: AGPLv3


src/SWP/Component/Common/Model/TimestampableTrait.php
```diff
@@ -26,6 +26,11 @@ trait TimestampableTrait
      */
     protected $updatedAt;
 
+    public function __construct()
+    {
+        $this->createdAt = new \DateTime();
```
ahilles107: it will be called also when class using it will have his own constructor?

takeit: when that Trait will be used by the class having its own constructor, the `createdAt` has to be set manually in that class.

When there is no constructor defined in the class, the value of `createdAt` will be set based on the Trait constructor.

PR #1023, opened 2020-01-07
add tests for article unpublishing and removing items from content list

ahilles107:
Makes sures that SWP-1841 is not valid

## Proposed Changes

  - add new mothod for easier restoring softdeleted entities

License: AGPLv3


src/SWP/Component/Common/Model/SoftDeletableTrait.php
```diff
@@ -15,32 +15,25 @@
...
-    /**
-     * @return bool
-     */
-    public function isDeleted()
+    public function isDeleted(): bool
     {
         return null !== $this->deletedAt;
     }
+
+    public function restore(): void
```
takeit: where is it used? I don't see it anywhere being used in this PR

PR #1025, opened 2020-01-07
Analytics export

takeit:
## Proposed Changes

This feature allows exporting analytics reports about the articles from a given time range to a CSV file via API. By default, it exports data from the last 30 days. 

The exported data are processed by the queue (RabbitMQ) and saved to the CSV file. Depends on the storage set in Publisher, the CSV file can be saved to AWS S3 or local storage.

When the report is ready, the email notification is sent to the user who requested such report, with the link to the CSV file.

There is also an API to list existing CSV reports so they can be downloaded at any time.

License: AGPLv3


src/SWP/Bundle/CoreBundle/AnalyticsExport/CsvFileWriter.php
```diff
@@ -0,0 +1,17 @@
+<?php
+
+namespace SWP\Bundle\CoreBundle\AnalyticsExport;
+
+class CsvFileWriter
+{
+    public function write(string $fileSourcePath, array $data): void
+    {
+        $fp = fopen($fileSourcePath, 'w');
```
ahilles107: cant we use here flysystem?

src/SWP/Bundle/CoreBundle/Controller/AnalyticsExportController.php
```diff
@@ -0,0 +1,161 @@
...
+     * @return SingleResourceResponse
+     *
+     * @throws \Exception
+     */
+    public function post(Request $request): SingleResourceResponse
+    {
+        /** @var UserInterface $currentlyLoggedInUser */
+        $currentlyLoggedInUser = $this->getUser();
+
+        $start = new DateTime($request->query->get('start', 'now'));
```
ahilles107: beside `$fileName` all those variables can be removed and values provided in `ExportAnalytics` constructor.

takeit: why?

ahilles107: they takes lines, and memory and can be safely provided directly in contructor (in one place)

takeit: but it's a micro-optimization which makes no sense here

src/SWP/Bundle/CoreBundle/AnalyticsExport/CsvFileWriter.php
takeit: I think it's an extra dependency which is not needed here because I can do the same using only three PHP built-in functions instead of injecting the whole service.

ahilles107: But we shouldn't write directly to filesystem. In multi server env or one like heroku it will fail. This file should be stored in some external storage (like s3) if it's configured.

takeit: it won't fail. It will save the file in the cache dir in this case (or any temp dir) and read it because it's the same process - at least on Heroku. It's just the temp file that is then uploaded to the S3 if configured. 

In this case, we have to change it in other places too. Because for example when downloading the author's avatar, generating webp format of images, article images, it's also writing to local temp storage and then uploads to s3 if configured or am I missing something?

ahilles107: ok, i was thinking that it's finall generated report destination - where link from email points.

takeit:
The failing test is not related to the added changes as it's failing in master.

PR #1031, opened 2020-01-09
update content list items based on list limit

takeit:
Fixes #
Content list items are not refreshed when the content list limit is changed.

## Proposed Changes
When the limit of the content list is changed, the items behind the set limit will be removed from the list.

License: AGPLv3


ahilles107:
Please check failed tests.

PR #1086, opened 2020-04-14
Redirect old articles slugs to an existing article

takeit:
## Proposed Changes

When the `override_slug_on_correction` tenant's setting is set to `true` and when the article's slugline is corrected (route or slug changes, to e.g. `/sports/football`), the old article's slug (e.g. `/business/article-1`) is stored in the article slug table.

When users visit `/business/article-1`, they will be redirected to `/sports/football` with status code 301.

When a slug is correct and the route is the same, e.g. `/business/article-1-edited`, then when users visit `/business/article-1`, they will be redirected to `/business/article-1-edited` with status code 301.

- store previous article slugs when the article's slug is corrected
- redirect old/previous article slugs to the existing article's URL

License: AGPLv3


src/SWP/Bundle/ContentBundle/EventListener/RedirectArticlePreviousSlugListener.php
```diff
@@ -0,0 +1,64 @@
...
+        $this->router = $router;
+        $this->articleSlugRepository = $articleSlugRepository;
+    }
+
+    public function onKernelException(ExceptionEvent $event): void
+    {
+        $exception = $event->getException();
+        $request = $event->getRequest();
+
+        if (!$exception instanceof NotFoundHttpException && !$exception instanceof ArticleNotFoundException) {
```
ahilles107, line 44: you can check only `ArticleNotFoundException` as it extends `NotFoundHttpException`

takeit:
> I have a problem with naming. `Slug` in this implementation is not a slug but a whole path - (route, slug, parameters etc). It can go like that, but maybe we can came up with some better name? It's visible especially on article entity where we have slug and slugs - and they keep different things.

I renamed it to `PreviousRelativeUrl`

PR #1087, opened 2020-04-20
add option to filter article metadata with expression

ahilles107:
Allow to catch published article to content list by expression filter.

```
{"metadata": "filter.contains('subject').containsItem('code', '001').containsItem('code', '123')"}
```

License: AGPLv3

### DIFF
--- a/src/SWP/Bundle/CoreBundle/Filter/DataFilter.php
+++ b/src/SWP/Bundle/CoreBundle/Filter/DataFilter.php
@@ -33,7 +33,7 @@ class DataFilter
         return $this;
     }
 
-    public function reset(): DataFilter
+    public function reset(): self
     {
         $this->currentData = $this->initialData;
     }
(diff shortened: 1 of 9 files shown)


src/SWP/Bundle/CoreBundle/Filter/DataFilter.php
```diff
@@ -0,0 +1,67 @@
...
+
+    public function loadData(array $data): self
+    {
+        $this->initialData = $data;
+        $this->currentData = $data;
+
+        return $this;
+    }
+
+    public function reset(): self
```
takeit, line 36: this function is not used anywhere

ahilles107: it's public one - so it's part of the api. It can be used for multiple checks, as `contains` changes verified data.  Example:
```
filter.contains('subject')
  .containsItem('code', '001')
  .containsItem('code', '123')
.reset()
.contains('custom_metadata')
  .containsItem('a', 'b')
```

filter returns true (match) as long as all checks don't throw exception.

PR #1093, opened 2020-04-28
pin articles on any position in content lists

takeit:
## Proposed Changes

To allow pinning of articles in both automatic and manual lists on any position. User can pin or unpin article in any position. The number of articles that could be pinned in the lists should not be limited.

License: AGPLv3

### DIFF
--- a/features/content_lists/pin_content_list_item_on_specific_position_in_manual_list.feature
+++ b/features/content_lists/pin_content_list_item_on_specific_position_in_manual_list.feature
@@ -1,0 +1,410 @@
+@disable-fixtures
+@content_lists
+Feature: Pin/unpin article in manual content list on any position
+  In order to pin article on specific position in the manual content list
+  As a HTTP Client
+  I want to be able to set the position of the pinned article
+
+  Scenario: Pin articles on specific position in the manual content list
+    Given the following Tenants:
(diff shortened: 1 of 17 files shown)


features/content_lists/pin_content_list_item_on_specific_position_in_manual_list.feature
```diff
@@ -0,0 +1,410 @@
...
+    Given I am authenticated as "test.user"
+    When I add "Content-Type" header equal to "application/json"
+    And I send a "PATCH" request to "/api/v2/content/lists/1/items/" with body:
+    """
+    {
+       "items":[
+          {
+             "content_id":1,
+             "action":"move",
+             "position":1,
```
ahilles107, line 366: maybe we should use here also `stickyPosition` to be consistent with manual lists?

takeit: this is the position you want to place the item on when it comes to the manual content list, the sticky position is a bit different because it applies only to the automatic lists and there we don't allow changing the position of other items although the position of items is saved. You can change the position of an item in an automatic content list only when you want to pin an article on the specific position so we have to keep that value somewhere thus the sticky position

PR #1099, opened 2020-04-29
handle migrated article orginal url and redirect it to new article location

ahilles107:
Fixes SWP-1978
### Add an option to save before migration url to redirects table

License: AGPLv3

### DIFF
--- a/features/content_push/migrated_articles_redirects.feature
+++ b/features/content_push/migrated_articles_redirects.feature
@@ -1,0 +1,51 @@
+Feature: Handle migrated articles redirects
+
+  Scenario: Push content and create redirect automatically to previous article url
+    Given I am authenticated as "test.user"
+    And I add "Content-Type" header equal to "application/json"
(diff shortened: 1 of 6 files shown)


features/content_push/migrated_articles_redirects.feature
```diff
@@ -0,0 +1,51 @@
...
+      "language":"en",
+      "body_html":"<p>some html body</p>",
+      "versioncreated":"2016-09-23T13:57:28+0000",
+      "firstcreated":"2016-09-23T09:11:28+0000",
+      "description_text":"some abstract text",
+      "version":"2",
+      "byline":"ADmin",
+      "keywords":[],
+      "extra":{
+        "original_article_url": "https://example.com/abstract-html-test"
```
takeit, line 18: It would be nice to document it somewhere that when the `original_article_url` is added to extra it will make a redirect or make it configurable via env var at least

PR #1114, opened 2020-05-20
move api docs to openapi3 format and remove nelmio api bundle annotations from controllers

ahilles107:
Remove NelmioApiBundle requirement from project. 
Remove all Swagger php annotiations from controllers. 
Put generated openapi3 docs to separate directory. 

It allows for not php developers to maintain documentation (improve it). 
It also removes twig and twig bundle dependency from bundles which provides only JSON responses. 

License: AGPLv3


takeit:
what about keeping it in a single openapi format file instead of multiple files?

It makes no sense to keep files just like this one which refers to the schema.

```
content:
  application/json:
    schema:
      $ref: ../schemas/WebhookType.yaml
```

takeit:
where would we host it now?

ahilles107:
It's also an option but then it will be one huge file, possibly harder to navigate through it and harder to review changes. We would host it in special controller (probably still on /api/doc) all what you need to load it is to open one index file.

takeit (approved):

PR #1119, opened 2020-05-25
Redirect to dynamic route after successful login

### LINKED ISSUE #1115: User is not redirected back to previous route after following successful OAuth flow
## Expected Behavior
As a user, I expect to be redirected to the article I was reading after a successful login via `/connect/oauth`.

## Current Behavior
I'm redirected to the homepage `/` after logging in via `/connect/oauth`.

## Possible Solution
Can we save the current user route when user initiates the OAuth flow in `/connect/oauth`? And then redirect back to it after a successful login at `/connect/oauth/check`. Right now, the OAuthController always redirects to the "homepage" route after a successful login:

https://github.com/superdesk/web-publisher/blob/6a36074e524f94046585a328237069e2d6698e39/src/SWP/Bundle/CoreBundle/Controller/ExternalOauthController.php#L31-L50

## Steps to Reproduce

1. Navigate to route other than homepage
2. Click on a link that navigates to `/connect/oauth`
3. Follow the oauth flow

kottkrig:
Fixes #1115 

## Proposed Changes

  - Start a session to keep track of a route that a user can be redirected to after a successful login

## Why was this PR needed?

Currently, after a successful login, the user is always redirected to the 'homepage' route. It would be nice if users were redirected back to an article if they started the login flow from an article.

## How can I test this PR?

Create a link to the oauth route with the query param `redirect_to_route_when_successful`.

```twig
<a href="{{ path('connect_oauth_start', {'redirect_to_route_when_successful': path(gimme.article)}) }}">Login</a></p>
```

License: AGPLv3

### DIFF
--- a/src/SWP/Bundle/CoreBundle/Controller/ExternalOauthController.php
+++ b/src/SWP/Bundle/CoreBundle/Controller/ExternalOauthController.php
@@ -7,18 +7,28 @@ namespace SWP\Bundle\CoreBundle\Controller;
 use Lexik\Bundle\JWTAuthenticationBundle\Services\JWTTokenManagerInterface;
 use Symfony\Component\HttpFoundation\Cookie;
 use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Session\SessionInterface;
 use Symfony\Bundle\FrameworkBundle\Controller\Controller;
 use Symfony\Component\Routing\Annotation\Route;
 
 class ExternalOauthController extends Controller
 {
     public const PUBLISHER_JWT_COOKIE = 'publisher_jwt';
+    public const PUBLISHER_REDIRECT_TO_AFTER_OAUTH = 'publisher_redirect_to_after_oauth';
 
     /**
      * @Route("/connect/oauth", name="connect_oauth_start")
      */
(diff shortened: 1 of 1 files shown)


kottkrig:
@takeit I've updated this PR to use the OAuth state to store the redirect-url instead of a new session.

Also, I now use the referer header as the redirect uri instead of using an explicitly set query variable.

src/SWP/Bundle/CoreBundle/Controller/ExternalOauthController.php
```diff
@@ -17,23 +18,25 @@ class ExternalOauthController extends Controller
...
+        $referer = $request->headers->get('referer');
+
         $clientRegistry = $this->get('knpu.oauth2.registry');
 
         return $clientRegistry
             ->getClient('external_oauth')
             ->redirect([
                 'openid', 'email', 'profile',
-            ]);
+            ], ["state" => $referer]);
```
takeit, line 31: @kottkrig please use single quotes here as we use them everywhere so I can merge that PR. Besides that, the PR LGTM 👍

kottkrig: Sure thing!

takeit:
Thanks, @kottkrig !

PR #1150, opened 2020-09-23
Improve full text search ES queries

takeit:
## Proposed Changes

  - improved full-text search
  - boost fields in order (the first field has the highest priority): title, lead, body, keyword, extra fields
  - boost author names
  - added API endpoint to search authors by phrase
  - filter packages and articles by author id instead of the author name
  - first, sort by relevance, then by e.g. custom sorting - `publishedAt` etc.
  - boost newer articles using decay functions. Articles published within last 1 month will rank higher, then the ones published within last year, and the rest
  - expose `maxScore` property in twig to determine if it's author or article search

After deploy:
- reindex elastic search

License: AGPLv3


src/SWP/Bundle/CoreBundle/Form/DataTransformer/AuthorToIdDataTransformer.php
```diff
@@ -0,0 +1,59 @@
+<?php
```
greg606, line 1: could we also use declare(strict_types=1); ?

src/SWP/Bundle/CoreBundle/Form/DataTransformer/AuthorToIdDataTransformer.php
```diff
@@ -0,0 +1,59 @@
...
+
+/*
+ * This file is part of the Superdesk Web Publisher Content Bundle.
```
greg606, line 4: this core bundle

src/SWP/Bundle/CoreBundle/Form/Type/AuthorSelectorType.php
```diff
@@ -0,0 +1,41 @@
+<?php
```
greg606, line 1: copyright missing

src/SWP/Bundle/ElasticSearchBundle/Controller/Api/AuthorSearchController.php
```diff
@@ -0,0 +1,74 @@
...
+namespace SWP\Bundle\ElasticSearchBundle\Controller\Api;
+
+use SWP\Bundle\ElasticSearchBundle\Criteria\Criteria;
+use SWP\Component\Common\Response\ResourcesListResponse;
+use SWP\Component\Common\Response\ResponseContext;
+use Symfony\Bundle\FrameworkBundle\Controller\Controller;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\Routing\Annotation\Route;
+
+class AuthorSearchController extends Controller
```
greg606, line 26: this will be a part of SF 5 migration work - remove deprecated controller and use injection for services

src/SWP/Bundle/ElasticSearchBundle/Criteria/Order.php
```diff
@@ -47,8 +47,6 @@ private function __construct($field, $direction)
     }
 
     /**
-     * @param array $parameters
-     *
      * @return Order
      */
     public static function fromQueryParameters(array $parameters)
```
greg606, line 52: missing return type

greg606 (approved):
ElasticSearchBundle\Repository\ArticleRepository::findByCriteria got really long with duplicated code in PackageRepository. Maybe it could be split into methods and duplication extracted to class? Just idea.

src/SWP/Bundle/ElasticSearchBundle/Controller/Api/AuthorSearchController.php
takeit: let's keep it as is for now and then do it properly in another PR as this is not related here right now

takeit:
We could split it but I would opt-in for keeping it as is for now because these are independent parts of code, used by different searches which vary a bit.

PR #1152, opened 2020-09-24
SWP-1977 - change extra from serialized string to relations 

greg606:
Fixes #
SWP-1977

## Proposed Changes

Adds Article relation Extrafields replacing serialized extra field
Based on change to metadata field here: https://github.com/superdesk/web-publisher/pull/1123

License: AGPLv3


src/SWP/Bundle/ContentBundle/Model/Article.php
```diff
@@ -446,4 +481,71 @@ public function setMetadata(array $metadata): void
...
+        }
+
+        foreach ($this->getExtraEmbedFields() as $extra) {
+            $this->removeExtraEmbedFields($extra);
+        }
+
+        foreach ($extra as $key => $value) {
+
+            if(is_array($value)) {
+                $embed = new ArticleExtraEmbedField();
```
takeit, line 538: You can create `named constructors` in `ArticleExtraEmbedField` and `ArticleExtraTextField` classes, so this part of the code is more readable, and we will have a one way to construct those objects.

takeit (changes requested):
Please, also run the PHP CS fixer on files you made changes to

takeit (approved):

src/SWP/Bundle/ContentBundle/Resources/config/serializer/Model.Article.yml
```diff
@@ -1,5 +1,7 @@
 SWP\Bundle\ContentBundle\Model\Article:
     exclusion_policy: ALL
+    accessor_order: custom
+    custom_accessor_order: [id,title,body,slug,publishedAt,status,route,isPublishable,metadata,lead,code,sources,extra,slideshows,previousRelativeUrls,createdAt,updatedAt,authors,keywords,media,templateName,publishStartDate,publishEndDate,featureMedia,seoMetadata]
```
ahilles107, line 4: why You want to maintain that order?

src/SWP/Bundle/CoreBundle/Migrations/2020/09/Version20200923162328.php
```diff
@@ -0,0 +1,119 @@
...
+        $this->abortIf(
+            'postgresql' !== $this->connection->getDatabasePlatform()->getName(),
+            'Migration can only be executed safely on \'postgresql\'.'
+        );
+
+        $this->addSql('DROP SEQUENCE swp_article_extra_id_seq CASCADE');
+        $this->addSql('DROP TABLE swp_article_extra');
+    }
+
+    public function postUp(Schema $schema): void
```
ahilles107, line 60: imo there it should be special command.

src/SWP/Bundle/ContentBundle/Resources/config/serializer/Model.Article.yml
greg606: because of behat tests requiring exact json content.

greg606: probably would make more sense to improve these tests somehow?

ahilles107: Yep, instead checking whole json we should check specific properties and their values.

takeit:
@greg606 please take a look at the comments, let's merge it to `master` branch and not `2.1` branch.

PR #1154, opened 2020-09-29
Improved performance of the metadata migration

takeit:
## Fixes

Improve the way the old metadata is converted to the new schema. The existing solution caused a memory leak.

## Proposed Changes

  - Improved queries for fetching old metadata in the migration.

License: AGPLv3


takeit:
failing tests are not related to this PR. Tests will be fixed in another PR.

PR #1164, opened 2020-11-03
Composer 2

greg606:
composer 2.0 stable has been released. It is advised to make project use it.

License: AGPLv3


ahilles107:
but remember that after that you will need to use composer v2 also in all deployments scripts

takeit (approved):

PR #1165, opened 2020-11-05
add inline ads feature

mehdignu:
Fixes #

## Proposed Changes

For the ad integration on articles we need a dynamic functionality. The ad integration should be depending on paragraphs and the numbers of characters.

The following rules should be respected:

- an adslots should be integrated after every paragraph (with 200 characters)
- an paragraph should have at least 200 characters
- an adslot should only be integrated if another paragraph follows
- no integration before and after one single paragraph
- a item of a list could be a paragraph if it has more than 200 characters
- no integration before a list
- no integration after headlines
- the number of characters should be adjustable in den backend
- has a paragraph less than 200 characters the next paragraph(s) should be added till the numbers of 200 characters is reached

License: AGPLv3


src/SWP/Bundle/CoreBundle/Twig/ArticleBodyExtension.php
```diff
@@ -49,9 +49,70 @@ public function getFunctions(): array
...
 
+    public function setInlineAds(string $body): array
+    {
+        # split the body into paragraphs
+        $paragraphs = explode("</p>", $body);
+
+        $tmp = [];
+        $splittedBody = [];
+        $parLength = 0;
+        $paragraphChars = $_ENV['ADS_CHARACTERS'];
```
ahilles107, line 64: You can't do this like that. Publisher (and symfony) have other way for accessing env variables. It should be passed to service via DI.

ahilles107:
where we can find tests?

.env
```diff
@@ -124,3 +124,6 @@ MAILER_DSN=smtp://localhost
 ###< symfony/mailer ###
 
 FROM_EMAIL=contact@publisher.test
+
+# after how many charachters in the body's paragraph the ad should to be placed
```
takeit, line 128: there is a typo in `charachters`

src/SWP/Bundle/CoreBundle/Tests/Twig/InlineAdsTest.php
```diff
@@ -0,0 +1,56 @@
...
+use SWP\Bundle\FixturesBundle\WebTestCase;
+
+class InlineAdsTest extends WebTestCase
+{
+    /**
+     * @var \Twig_Environment
+     */
+    private $twig;
+
+    public function setUp()
```
takeit, line 26: please add `:void` return type to all the functions in this class and `declare(strict_types=1);` after `<?php`

takeit (changes requested):
@mehdignu please see my comments and we are ready to merge. The PR LGTM.

PR #1168, opened 2020-11-26
Attach article to content route on article publish

greg606:
Fixes #
When publishing an article to content type of route, we also need to assign its id to route's 'content' property. 
Only then that article appears on frontend as content of that route.

License: AGPLv3

### DIFF
--- a/src/SWP/Bundle/ContentBundle/EventListener/AttachArticleToContentRouteListener.php
+++ b/src/SWP/Bundle/ContentBundle/EventListener/AttachArticleToContentRouteListener.php
@@ -26,7 +26,7 @@ class AttachArticleToContentRouteListener
         $article = $articleEvent->getArticle();
         $route = $articleEvent->getArticle()->getRoute();
 
-        if(RouteInterface::TYPE_CONTENT === $route->getType()) {
+        if($route && RouteInterface::TYPE_CONTENT === $route->getType()) {
             $route->setContent($article);
             $this->routeRepository->persist($route);
             $this->routeRepository->flush();
@@ -39,7 +39,7 @@ class AttachArticleToContentRouteListener
 
         $route = $articleEvent->getArticle()->getRoute();
 
-        if(RouteInterface::TYPE_CONTENT === $route->getType()) {
+        if($route && RouteInterface::TYPE_CONTENT === $route->getType()) {
             $route->setContent(null);
             $this->routeRepository->persist($route);
             $this->routeRepository->flush();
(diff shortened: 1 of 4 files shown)


src/SWP/Bundle/ContentBundle/EventListener/AttachArticleToContentRouteListener.php
```diff
@@ -0,0 +1,51 @@
...
+    {
+
+        $this->routeRepository = $routeRepository;
+    }
+
+    public function onArticlePublish(ArticleEvent $articleEvent): void
+    {
+        /** @var ArticleInterface $article */
+        $article = $articleEvent->getArticle();
+        $route = $articleEvent->getArticle()->getRoute();
```
takeit, line 27: You don't need to do `$articleEvent->getArticle()`, you can just use $article straight away here

src/SWP/Bundle/ContentBundle/EventListener/AttachArticleToContentRouteListener.php
```diff
@@ -0,0 +1,51 @@
...
+        $alreadyAttachedRoute = $this->routeRepository->findOneBy(['content' => $article]);
+
+        if($route && !$alreadyAttachedRoute &&  RouteInterface::TYPE_CONTENT === $route->getType()) {
```
takeit, line 30: I would make this if more strict: `null !== $route` instead of just `$route`

src/SWP/Bundle/ContentBundle/EventListener/AttachArticleToContentRouteListener.php
```diff
@@ -0,0 +1,51 @@
...
+            $route->setContent($article);
+            $this->routeRepository->persist($route);
```
takeit, line 32: we don't need to persist the route here as we are updating it. The flush would be enough

src/SWP/Bundle/ContentBundle/EventListener/AttachArticleToContentRouteListener.php
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    public function onArticleUnpublish(ArticleEvent $articleEvent): void
+    {
+
+        $route = $articleEvent->getArticle()->getRoute();
+
+        if($route && RouteInterface::TYPE_CONTENT === $route->getType()) {
+            $route->setContent(null);
+            $this->routeRepository->persist($route);
```
takeit, line 45: same here, persist is not needed

takeit (changes requested):
Please fix the code style of PHP classes using PHP cs fixer

takeit:
Please also add the behat tests to test that use case.

takeit (approved):

PR #1169, opened 2020-11-30
added byline to the analytics export

takeit:
## Proposed Changes

  - show `byline` field in the analytics reports

License: AGPLv3


takeit:
The failing tests are not related to those changes. It will be fixed in another PR.

PR #1170, opened 2020-12-08
set x-forwarded-for from x-real-ip header

takeit:
## Reasons
Properly detect client real IP in Publisher for the geolocation feature. When the app is behind the proxy the real client IP address is the IP of the proxy.

## Proposed Changes

  - set value of `x-forwarded-for` header from `x-real-ip` header

License: AGPLv3

### DIFF
--- a/public/index.php
+++ b/public/index.php
@@ -19,6 +19,8 @@ if ($_SERVER['APP_DEBUG']) {
     Debug::enable();
 }
 
+$_SERVER['HTTP_X_FORWARDED_FOR'] = $_SERVER['HTTP_X_REAL_IP'];
+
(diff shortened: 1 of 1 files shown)


takeit:
Failed tests are not related to these changes. The tests will be fixed in another PR.

public/index.php
```diff
@@ -19,6 +19,8 @@
     Debug::enable();
 }
 
+$_SERVER['HTTP_X_FORWARDED_FOR'] = $_SERVER['HTTP_X_REAL_IP'];
```
ahilles107, line 22: What if it will be not set and `HTTP_X_FORWARDED_FOR` will have real value? As it can be populated by nginx config rules? I would add some checks, but best would be to handle this on nginx level - not here.

PR #1187, opened 2021-02-16
expose search articles api endpoint

greg606:
## Reasons

For PWA applications the search is required to be exposed without jwt authentication

License: AGPLv3


takeit:
it should go to master

PR #1198, opened 2021-03-22
Add PWA config to tenant

greg606:
## Reasons

To allow proper operation of PWA enabled tenant. 

## Proposed Changes

License: AGPLv3


features/tenants/creating_a_new_tenant_with_pwa_config.feature
```diff
@@ -0,0 +1,40 @@
...
+    Given I am authenticated as "test.user"
+    When I add "Content-Type" header equal to "application/json"
+    And I send a "POST" request to "/api/v2/tenants/" with body:
+     """
+      {
+          "domain_name": "example.com",
+          "name": "Example tenant",
+          "subdomain": "tenant1",
+          "pwa_config": {
+            "url": "pwa_url"
```
takeit, line 16: it should be validated so it's possible to enter only enter valid urls

PR #1212, opened 2021-10-29
Convert image palette to truecolor before converting to webp

kottkrig:
## Reasons

`ImageConversionHandler.php` would crash on images in the color space _Adobe RGB (1998)_ with the error: `Paletter image not supported by webp`.

## Proposed Changes

Normalize the color space to truecolor before converting it to webp.

License: AGPLv3


takeit:
Thanks @kottkrig !

PR #1214, opened 2021-11-02
Urlencode subscription cache key

kottkrig:
## Reasons

`CachedSubscriptionsProvider` will throw an error when the generated cache key contains a space.

```
Doctrine\Common\Cache\InvalidCacheId: Cache id "/var/www/publisher[subscription_provider_xxxxtest@example.com__NAME WITH SPACE][1]" contains unauthorized character " "
```

## Proposed Changes

Urlencode the generated cache key to ensure that it only contains valid characters.

License: AGPLv3


takeit:
Thanks @kottkrig !

PR #1217, opened 2022-01-20
Cancel article timestampable when updating mediaUpdatedAt

### LINKED ISSUE #1215: Bug: Article updatedAt field is updated when media is converted to webp
## Current behaviour

Currently, in `swp_article` table, the value in column `updated_at` is updated each time the article's media renditions are converted to webp. 

Here are some rows from our database where we can see that `updated_at` is set to the same timestamp as `media_updated_at`. These articles haven't been modified in any way other than the article images being converted to webp.

| id    | created_at          | updated_at          | media_updated_at    |
|-------|---------------------|---------------------|---------------------|
| 44152 | 2021-11-23 10:31:13 | 2021-11-24 16:05:22 | 2021-11-24 16:05:22 |
| 44151 | 2021-11-23 10:04:25 | 2021-11-24 16:02:49 | 2021-11-24 16:02:49 |
| 44150 | 2021-11-23 09:58:29 | 2021-11-24 16:02:40 | 2021-11-24 16:02:40 |
| 44149 | 2021-11-23 09:53:33 | 2021-11-24 16:02:16 | 2021-11-24 16:02:16 |
| 44148 | 2021-11-23 09:41:27 | 2021-11-24 16:02:07 | 2021-11-24 16:02:07 |
| 44147 | 2021-11-23 09:31:17 | 2021-11-24 16:01:53 | 2021-11-24 16:01:53 |

## Expected behaviour

Only the column `media_updated_at` should be updated on each webp conversion. `updated_at` should not be altered.

## Possible fix/solution?

We haven't been able to pinpoint the code that is responsible for the updated `updated_at` field. Is this perhaps an expected Symfony behaviour?

As far as we can see, the ImageConversionHandler seems to be doing the correct thing. It updates the `media_updated_at` field.

https://github.com/superdesk/web-publisher/blob/ef8afc6cdefab1b0fde38063c38eefd1c508590d/src/SWP/Bundle/CoreBundle/MessageHandler/ImageConversionHandler.php#L132-L139

kottkrig:
## Reasons

Currently, article "updatedAt" is updated when images are successfully converted to webp. The expected behaviour would be to only update the "mediaUpdatedAt" field. This PR fixes #1215.

## Proposed Changes

Call `TimestampableCancelInterface.cancelTimestampable(true)` to cancel the automatic update of "updatedAt" field on article.

License: AGPLv3

### DIFF
--- a/src/SWP/Bundle/CoreBundle/MessageHandler/ImageConversionHandler.php
+++ b/src/SWP/Bundle/CoreBundle/MessageHandler/ImageConversionHandler.php
@@ -134,7 +134,9 @@ class ImageConversionHandler implements MessageHandlerInterface
         /** @var ImageRenditionInterface[] $articleMedia */
         $articleMedia = $this->imageRenditionRepository->findBy(['image' => $image]);
         foreach ($articleMedia as $media) {
-            $media->getMedia()->getArticle()->setMediaUpdatedAt(new DateTime());
+            $article = $media->getMedia()->getArticle();
+            $article->setMediaUpdatedAt(new DateTime());
+            $article->cancelTimestampable(true);
         }
     }
 


takeit:
thank you, @kottkrig!

